Issue #148, opened 2022-10-20
Rewrite ETOS client to become more user-focused


t-persson:
### Description

The client was written by me and for me which is bad from a usability perspective. I want to add more and clearer information about what is going on when running ETOS.

### Motivation

Hard to understand what is going on when running ETOS.

### Exemplification

It would be easier to debug or know what to do when things go wrong.

### Benefits

UX

### Possible Drawbacks

None

PR #168, opened 2023-06-07
Logs from log proxy

t-persson:
### Applicable Issues

Depends-on: eiffel-community/etos-api#41
#148 

### Description of the Change

Connect to ETOS log proxy and get logs from running ETOS test runs.
For reviewing it is recommended to review the commits in isolation as the, now, Downloader was renamed from Logs to Downloader in order for this new module to be named Logs. The full diff looks weird :) 

### Alternate Designs

I decided to keep the old messages as they are, since they still keep some valuable information. They should be removed in the future (speaking specifically about the announcer).

### Benefits

This will allow ETOS client to give better information when certain ETOS components are stuck in internal work.

### Possible Drawbacks

Another dependency for logs and events. We should replace the event repository queries to use the ETOS API in the future.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/cli/src/etos_client/test_run/test_run.py
+++ b/cli/src/etos_client/test_run/test_run.py
@@ -1,0 +1,130 @@
+# Copyright Axis Communications AB.
+#
+# For a full list of individual contributors, please see the commit history.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


cli/src/etos_client/test_run/test_run.py
```diff
@@ -0,0 +1,130 @@
...
+        logger = getattr(self.remote_logger, message.level)
+        logger(
+            message,
+            extra={
+                "rname": message.name,
+                "rtime": message.datestring,
+            },
+        )
+
+    def __log_until_eof(self, etos: ETOS, timeout: int) -> Iterator[Ping]:
```
fredjn, line 122: It's not terribly obvious that the timeout is actually the end_time in this case. Perhaps rename the variable or document.

fredjn (changes requested):

fredjn (approved):

PR #237, opened 2024-05-29
Attempt to reduce the number of ER requests

t-persson:
### Applicable Issues

#238 

### Description of the Change

ETOS is doing a lot of requests to the event repositories. This is not scalable at all and we need to reduce the amount of queries we are doing and start working on a better solution.

### Alternate Designs

We could go to the final solution right away, but that is quite a ways away and we should try to reduce the queries now.

### Possible Drawbacks

We will be missing some test case information with this change, this is just flavor and not necessary for an ETOS testrun.
There will also be a few logs that may come at weird times such as "waiting for test suite to start" when you can clearly see from the SSE logs that the suite has started. This should be fine in my opinion.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/cli/src/etos_client/test_run/test_run.py
+++ b/cli/src/etos_client/test_run/test_run.py
@@ -118,7 +118,6 @@ class TestRun:
 
     def __announce(self, events: Events) -> None:
         """Announce current state of ETOS."""
-        # events = self.__collector.collect_test_case_events()
         self.__announcer.announce(events, self.__downloader.downloads)
 
     def __download(self, events: Events) -> None:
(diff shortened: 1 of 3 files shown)


cli/src/etos_client/test_run/test_run.py
```diff
@@ -74,13 +74,17 @@ def track(self, etos: ETOS, timeout: int) -> Events:
...
-        self.__last_log = time.time()
+        last_log = time.time()
         timer = None
         for _ in self.__log_until_eof(etos, end):
-            events = self.__collect(etos)
-            self.__status(events)
-            self.__announce()
-            self.__download(events)
+            if last_log + self.log_interval <= time.time():
+                events = self.__collect(etos)
```
t-persson, line 81: Wrong collector

t-persson: And the logic is not sound

fredjn (changes requested):

fredjn (approved):

Issue #238, opened 2024-06-04
ETOS should reduce the number of ER requests made


t-persson:
### Description

An ETOS testrun makes too many requests to the ER and this is very scalable. We shall reduce the number of requests

PR #239, opened 2024-06-04
Get logs from api

t-persson:
### Applicable Issues

fixes: #238 
based-on: #237 

### Description of the Change

Get URL:s for logs and artifacts from ETOS instead of Eiffel.
I also removed a bunch of unused events and the entire announcer module.

### Alternate Designs

We probably want to stream the log URLs using the SSE server in the future. I chose this way of doing it as it is much easier to verify and we are in a bit of a hurry to fix this. It is also a much simpler solution.

### Possible Drawbacks

A lot of traffic going to the new logarea API instead of the event repository but I do believe that it is much easier for us to scale this new API than the GraphQL API.

### Sign-off

    indicated in the file; or

    in the file; or

    it.


cli/src/etos_client/test_run/test_run.py
```diff
@@ -91,43 +105,50 @@ def track(self, etos: ETOS, timeout: int) -> Events:
...
-            events = self.__collector.collect_test_case_events()
-            self.__announcer.announce(events, self.__downloader.downloads)
-            self.__last_log = time.time()
-
-    def __download(self, events: Events) -> None:
+        if not events.tercc:
+            self.logger.info("Waiting for ETOS to start.")
+            return
+        if not events.activity.triggered:
+            self.logger.info("Waiting for ETOS to start.")
```
andmat900, line 135: Please make the error message different from line 132. It will be easier to understand in the logs where exactly waiting is happening.

t-persson: These messages are for the user, not for us. The difference between a Tercc or activity means nothing to them.
For us, we would see that this check is for either one of these two and could check the ER for more information.

t-persson: @andmat900

cli/src/etos_client/events/events.py
fredjn, line 1: The TestCase class is AFAICT unused.

fredjn (approved):

Issue #240, opened 2024-07-08
Enable automatic compaction of key revisions in ETCD


andmat900:
### Description

Currently, etcd is storing all key history and the database can grow indefinitely. To avoid this:

- Revision auto compaction shall be enabled for etcd pods
- Defragmentation shall be done on a scheduled basis
- 

### Expected Behavior

Only latest key revisions shall be stored.

### The version of this project/repo, if applicable

main

### The version/edition of the Eiffel Protocol used, if applicable

latest

andmat900:
Released in ETOS 6.6.2.

PR #241, opened 2024-07-08
Enable auto compaction for etcd

andmat900:
### Applicable Issues

- https://github.com/eiffel-community/etos/issues/240

### Description of the Change

This change enables automatic compaction for etcd.

### Sign-off

    indicated in the file; or

    in the file; or

    it.


manifests/etcd/statefulset.yaml
```diff
@@ -39,7 +39,9 @@ spec:
               --initial-cluster-token etcd-cluster-1 \
               --initial-cluster ${PEERS} \
               --initial-cluster-state new \
-              --data-dir /var/run/etcd/default.etcd
+              --data-dir /var/run/etcd/default.etcd \
+              --auto-compaction-mode=revision \
+              --auto-compaction-retention=1
```
t-persson, line 44: 1 hour seems extremely low when an ETOS testrun can run for about 24 hours.

manifests/etcd-defrag/cronjob.yaml
```diff
@@ -0,0 +1,21 @@
...
+metadata:
+  name: etcd-defrag
+spec:
+  schedule: "0 0 * * *"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+          - name: etcd-defrag
```
t-persson, line 12: Is defrag actually necessary? According to https://etcd.io/docs/v3.2/op-guide/maintenance/ :
`Compacting old revisions internally fragments etcd by leaving gaps in backend database. Fragmented space is available for use by etcd but unavailable to the host filesystem.`

manifests/etcd/statefulset.yaml
andmat900: The mode is set to "revision", so it is one revision, not hours.

manifests/etcd-defrag/cronjob.yaml
andmat900: Yes, correct. Without defragmentation the problem is the free space on the host filesystem.

> After compacting the keyspace, the backend database may exhibit internal fragmentation. Any internal fragmentation is space that is free to use by the backend but still consumes storage space. The process of defragmentation releases this storage space back to the file system.

So defragmentation is needed to free up the disk space on the host filesystem.

t-persson: But the host filesystem in this case is a kubernetes volume used exclusively by etcd. Why do we need to release it?

t-persson (approved):

Issue #242, opened 2024-07-15
ETOS to be run as a kubernetes operator instead


t-persson:
### Summary

**As** "a system engineer"
**I want** "better control of testruns"
**So that** "I can more properly manage the ETOS system."

### Context

ETOS has become quite complex and we are seeing a lot of issues with regards to how we handle environments and the traceability of ETOS in general.
We should off-load some of this complexity to Kubernetes instead by writing an ETOS operator/controller. This way we could remove some complexity and implement it in the Reconcile loop of ETOS instead of having services deal with it themselves.
We can also more easily track faulty environments and testruns and make sure we release environments whenever a custom environment resource is deleted.

PR #243, opened 2024-07-15
Initial implementation of an ETOS Kubernetes controller

t-persson:
### Applicable Issues

#242

### Description of the Change

This implements the controller for testrun, environment and providers. Environment is only used for checking that providers are available and the provider implementation will only check if external providers respond properly on a healthcheck endpoint.

The testrun controller can run a suite runner if an environment reports that it is ready and it will wait for the suite runner to complete and delete it and environment after execution. The testrun controller is quite hard-coded though and does not deploy the log listener sidecar or any logstash sidecars for the suite runner.

This change should be seen as a PoC and not to be used in production.

The change is two commits. First commit is boilerplate code that is created by kubebuilder, the second is my implementation.

### Alternate Designs

Too many to list. I believe a discussion will be required instead of me listing a ton of stuff here.

### Possible Drawbacks

I don't actually see any real drawbacks. This is just better than what we have now.

### Sign-off

    indicated in the file; or

    in the file; or

    it.


internal/controller/environment_controller.go
```diff
@@ -0,0 +1,138 @@
...
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.18.4/pkg/reconcile
+func (r *EnvironmentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+
+	// Get environment if exists.
+	environment := &etosv1alpha1.Environment{}
+	err := r.Get(ctx, req.NamespacedName, environment)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			logger.Info("environment not found. ignoring object")
```
andmat900, line 68: General question: is there enough info logging? I guess observability is important for controllers in general, we want to be able to see the state.

internal/controller/environment_controller_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+var _ = Describe("Environment Controller", func() {
+	Context("When reconciling a resource", func() {
+		const resourceName = "test-resource"
+
+		ctx := context.Background()
+
+		typeNamespacedName := types.NamespacedName{
+			Name:      resourceName,
+			Namespace: "default", // TODO(user):Modify as needed
```
andmat900, line 41: Shall it be hard-coded? Infer from environment?

internal/controller/provider_controller.go
```diff
@@ -0,0 +1,104 @@
...
+		if apierrors.IsNotFound(err) {
+			logger.Info("provider not found. ignoring object")
+			return ctrl.Result{}, nil
+		}
+		logger.Error(err, "failed to get provider")
+		return ctrl.Result{}, err
+	}
+	logger.V(2).Info("Checking availability of provider", "provider", req.NamespacedName)
+
+	// TODO: Schedule checks instead of doing it every time something happens
```
andmat900, line 66: Write an issue?

internal/controller/provider_controller.go
```diff
@@ -0,0 +1,104 @@
...
+
+	// We don't check the availability of JSONTas as it is not yet running as a service we can check.
+	if (etosv1alpha1.JSONTas{}) == provider.Spec.JSONTas {
+		resp, err := http.Get(fmt.Sprintf("%s/%s", provider.Spec.Host, provider.Spec.Healthcheck.Endpoint))
+		if err != nil {
```
andmat900, line 71: Should retries be done here? Or is it handled by Kubernetes outside this controller?

internal/controller/testrun_controller.go
```diff
@@ -0,0 +1,405 @@
...
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.18.4/pkg/reconcile
+func (r *TestRunReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+
+	logger.V(2).Info("Get testrun", "namespace", req.Namespace, "name", req.Name)
+	testrun := &etosv1alpha1.TestRun{}
+	if err := r.Get(ctx, req.NamespacedName, testrun); err != nil {
+		if apierrors.IsNotFound(err) {
+			logger.Info("Testrun not found, exiting", "namespace", req.Namespace, "name", req.Name)
```
andmat900, line 73: The error message seems to be the same for both scenarios:
- when the test run doesn't exist
- when there was an error getting the test run
Maybe use different error message.

internal/controller/testrun_controller.go
```diff
@@ -0,0 +1,405 @@
...
+			return ctrl.Result{}, nil
+		}
+		logger.Error(err, "Testrun not found", "namespace", req.Namespace, "name", req.Name)
+		return ctrl.Result{}, err
+	}
+
+	requeue, err := r.reconcile(ctx, testrun, req)
+	if err != nil {
+		if errors.IsConflict(err) {
+			return ctrl.Result{Requeue: true}, nil
```
andmat900, line 83: This assumes that any conflict error shall be solved by requeue. Is it correct?

internal/controller/testrun_controller.go
```diff
@@ -0,0 +1,405 @@
...
+			logger.Info("Setting suiterunner (success) false")
+			return false, r.Status().Patch(ctx, testrun, patch)
+		}
+	}
+	return false, nil
+}
+
+func (r *TestRunReconciler) getOrCreateSuiteRunner(ctx context.Context, testrun *etosv1alpha1.TestRun, name types.NamespacedName) (*batchv1.Job, error) {
+	suiteRunner := &batchv1.Job{}
+	if err := r.Get(ctx, name, suiteRunner); err != nil {
```
andmat900, line 217: If Get fails, is some logging needed?

internal/controller/testrun_controller.go
```diff
@@ -0,0 +1,405 @@
...
+			return suiteRunner, err
+		}
+	}
+	return suiteRunner, nil
+}
+
+// getOrCreateEnvironment attempts to get an environment or create a new one if it does not exist.
+func (r *TestRunReconciler) getOrCreateEnvironment(ctx context.Context, testrun *etosv1alpha1.TestRun, name types.NamespacedName) (*etosv1alpha1.Environment, error) {
+	environment := &etosv1alpha1.Environment{}
+	if err := r.Get(ctx, name, environment); err != nil {
```
andmat900, line 240: Logging if get fails?

internal/controller/testrun_controller.go
```diff
@@ -0,0 +1,405 @@
...
+		if condition.Type == conditionType {
+			return condition.Status == status
+		}
+	}
+	return false
+}
+
+// suiteRunnerJob is the job definition for an etos suite runner.
+func (r TestRunReconciler) suiteRunnerJob(tercc []byte, testrun *etosv1alpha1.TestRun) *batchv1.Job {
+	ttl := int32(300)
```
andmat900, line 291: Make (some of) these values configurable?

internal/controller/testrun_controller.go
```diff
@@ -0,0 +1,405 @@
...
+	grace := int64(30)
+	backoff := int32(0)
+	job := &batchv1.Job{
```
andmat900, line 294: Is there enough validation for the values? For example, a check that testrun.Spec.SuiteRunnerImage is valid? 

If Job instance fails to initialize, should there be some error handling?

internal/controller/testrun_controller_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+var _ = Describe("TestRun Controller", func() {
+	Context("When reconciling a resource", func() {
+		const resourceName = "test-resource"
+
+		ctx := context.Background()
+
+		typeNamespacedName := types.NamespacedName{
+			Name:      resourceName,
+			Namespace: "default", // TODO(user):Modify as needed
```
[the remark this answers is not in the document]
t-persson: This is a test. It's fine that it is hard-coded.

internal/controller/testrun_controller_test.go
```diff
@@ -0,0 +1,84 @@
...
+		BeforeEach(func() {
+			By("creating the custom resource for the Kind TestRun")
+			err := k8sClient.Get(ctx, typeNamespacedName, testrun)
+			if err != nil && errors.IsNotFound(err) {
+				resource := &etosv1alpha1.TestRun{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      resourceName,
+						Namespace: "default",
+					},
+					// TODO(user): Specify other spec details if needed.
```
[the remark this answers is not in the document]
t-persson: `TODO(user)` can be ignored.

internal/controller/environment_controller.go
```diff
@@ -0,0 +1,138 @@
...
+			return ctrl.Result{}, err
+		}
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// checkProviders checks if all providers for this environment are available.
+func (r EnvironmentReconciler) checkProviders(ctx context.Context, environment *etosv1alpha1.Environment) error {
+	err := r.checkProvider(ctx, environment.Spec.IUTProvider, environment.Namespace, &etosv1alpha1.Provider{})
```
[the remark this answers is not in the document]
t-persson: I think that would reduce visibility

internal/controller/environment_controller.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		return err
+	}
+	err = r.checkProvider(ctx, environment.Spec.ExecutionSpaceProvider, environment.Namespace, &etosv1alpha1.Provider{})
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
t-persson: If one fails, then we cannot use the environment. We only report `Ready` and `Status` for an Environment.
If we were to report `IUTProviderStatus`, `LogAreapProviderStatus` and `ExecutionSpaceProviderStatus`, then I agree with it. I won't do that in this PR though.

internal/controller/environment_controller.go
t-persson: There's never enough logging :) For now I think it's good enough. We need to identify which information is important to log and then make sure we log enough, but not too much.
For the line that you posted this comment, that is a message that gets logged if the `Environment` is deleted, I.e. using `kubectl delete environment/name`.

internal/controller/environment_controller_test.go
t-persson: It's fine to hard-code this in tests.

internal/controller/provider_controller.go
t-persson: Yes. After merge

internal/controller/provider_controller.go
t-persson: This Reconciliation loop is run, at least, every `provider.Spec.Healthcheck.IntervalSeconds`(default 30) seconds.

internal/controller/testrun_controller.go
t-persson: If the testrun does not exist, we cannot do anything. This log message is printed only if the testrun has been deleted.
Outside of the `apierrors.IsNotFound(err)` (line 76) we log the error we got back from `r.Get`

internal/controller/testrun_controller.go
t-persson: It is correct.
Because of how controller operate, the `status` field may be updated by other controllers. If the `status` field has been updated by another we will get a conflict. That is what this check is catching. A requeue is the proper way to handle it.

internal/controller/testrun_controller.go
t-persson: I agree that having them hard-coded like this is sub-optimal however I do not want to provide them to the users either, which wee would do if we add it to the TestRun spec.
I suggest leaving it for now and see if we can find a better solution later.

internal/controller/testrun_controller.go
t-persson: This would be caught on line 199 `case batchv1.JobFailed`

t-persson: Done

andmat900 (approved):

PR #244, opened 2024-07-19
Etos cluster controller

t-persson:
### Applicable Issues

#242 

### Description of the Change

The cluster resource will create all the services needed for an ETOS deployment, including extra infrastructure needs if needed. It is designed to make it a lot easier to get an ETOS cluster up and running, especially when running ETOS in a local kubernetes cluster.
Note that this cluster resource is not complete yet. It does not, for example, deploy any providers for an ETOS testrun.
It does not read secrets that can be added in the cluster custom resource.
It does not deploy a logstash instance for ETOS logging.

### Alternate Designs

I wanted to create a way for developers to easily deploy a test cluster and decided to see how easy it was to create a controller for it. Another way could be to deploy all services separately and let ETOS deploy ETOS and nothing else, but we still need to configure ETOS with the information on how to communicate with the extra services so I thought that it would just be easier for ETOS to deploy them.
I also created this custom resource with the ability to deploy a production cluster where all extra services are deployed elsewhere, but I have not verified this yet.

### Possible Drawbacks

It is an extra thing to maintain, but I think that having access to a local ETOS cluster makes it worth it and if we use this cluster for the production deployments as well, then the maintainer burden will be lessened.

### Sign-off

    indicated in the file; or

    in the file; or

    it.


internal/etos/api/logarea.go
```diff
@@ -0,0 +1,241 @@
...
+	Scheme *runtime.Scheme
+}
+
+// NewETOSLogAreaDeployment will create a new ETOS logarea reconciler.
+func NewETOSLogAreaDeployment(spec etosv1alpha1.ETOSLogArea, scheme *runtime.Scheme, client client.Client) (*ETOSLogAreaDeployment, error) {
+	return &ETOSLogAreaDeployment{spec, client, scheme}, nil
+}
+
+// Reconcile will reconcile the ETOS logarea to its expected state.
+func (r *ETOSLogAreaDeployment) Reconcile(ctx context.Context, cluster *etosv1alpha1.Cluster) error {
```
andmat900, line 54: General question, applies to all components: should there be more verbose info logging? Observability seems important here, could help a lot with debugging.

internal/etos/api/logarea.go
```diff
@@ -0,0 +1,241 @@
...
+			},
+		},
+		TimeoutSeconds:   1,
+		PeriodSeconds:    10,
+		SuccessThreshold: 1,
+		FailureThreshold: 3,
+	}
+	return corev1.Container{
+		Name:  name.Name,
+		Image: r.Image,
```
andmat900, line 198: Applies to all services: Should this include ImagePullPolicy setting? Partly, if we use mutable tags sometimes ("latest") or if we want to update images more easily (like in a dev environment).

internal/etos/database.go
```diff
@@ -0,0 +1,317 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+var (
+	etcdClientPort int32 = 2379
```
andmat900, line 37: applies to all services: is it possible to avoid hard-coding these parameters?

t-persson: Possible, but not necessary. Kubernetes won't get any port conflicts. I would rather use the default ports for the services unless there is a real problem with using it.

internal/etos/api/logarea.go
t-persson: We should need more yes. We should, however, identify the need first and add only relevant logs. The controller will easily get spammy.

internal/controller/cluster_controller_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+var _ = Describe("Cluster Controller", func() {
+	Context("When reconciling a resource", func() {
+		const resourceName = "test-resource"
+
+		ctx := context.Background()
+
+		typeNamespacedName := types.NamespacedName{
+			Name:      resourceName,
+			Namespace: "default", // TODO(user):Modify as needed
```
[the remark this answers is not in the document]
t-persson: No need to do that for the tests.

internal/controller/cluster_controller.go
```diff
@@ -0,0 +1,168 @@
...
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.18.4/pkg/reconcile
+func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+
+	// TODO: Logstash
```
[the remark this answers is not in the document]
t-persson: Yes. After merge

internal/etos/api/logarea.go
t-persson: Added to services where we allow image to be changed.

andmat900 (approved):

PR #252, opened 2024-08-13
Log the correct fail/error message

fredjn:
### Description of the Change

Since we no longer collect information/events on subsuites
the logic fo logging messages about failures in sub suites
was not workin properly. This resulted in incorrect error
message when a suite successfully executed but had test case
failures in it.

### Possible Drawbacks
Log message will be a tad less detailed in it's information (e.g no information about the number of sub suites)

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by: Fredrik Fristedt <<fredrik.fristedt@axis.com>>

### DIFF
--- a/cli/src/etos_client/test_results/test_results.py
+++ b/cli/src/etos_client/test_results/test_results.py
@@ -57,12 +57,6 @@ class TestResults:
         """Build test results based on events retrieved."""
         if not self.__has_failed(test_suites):
             return True, "Test suite finished successfully."
-
-        failures = 0
-        sub_suites = 0
-        for test_suite in test_suites:
-            failures += self.__count_sub_suite_failures(test_suite.sub_suites)
-            sub_suites += len(test_suite.sub_suites)
         messages = self.__fail_messages(test_suites)
         if len(messages) == 1:
             return False, messages[0]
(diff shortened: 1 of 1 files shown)


t-persson (approved):

cli/src/etos_client/test_results/test_results.py
```diff
@@ -57,22 +57,14 @@ def __test_result(self, test_suites: list[TestSuite]) -> tuple[bool, str]:
...
         if len(messages) == 1:
             return False, messages[0]
         if messages:
             for message in messages[:-1]:
                 self.logger.error(message)
             return False, messages[-1]
-        if sub_suites == 0:
-            return False, "ETOS failed to start any test suites"
-        return False, f"{failures}/{sub_suites} test suites failed."
+        return False, f"Test case failures during test suite execution"
```
andmat900, line 67: Just to double-check, ignore if it is out-of-scope:

If the suite JSON file contains an empty list, will this be handled by this method? I remember I saw weird error messages in that case before.

fredjn: I added a check for empty suite definition list, and I also checked some other sad path cases with broken suite definitions. It seems to work sas expected.

PR #259, opened 2024-08-29
Enable downloading from internal report and artifact events

t-persson:
### Description of the Change

Listen for Report and Artifact events instead of polling the log area server when downloading files.

### Alternate Designs

None

### Possible Drawbacks

None

### Sign-off

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/cli/src/etos_client/downloader/downloader.py
+++ b/cli/src/etos_client/downloader/downloader.py
@@ -56,6 +56,28 @@ class Downloadable(BaseModel):
     path: Path = Path.cwd()
 
 
+class IntegrityError(Exception):
+    """Integrity verification error."""
+
+    def __init__(self, url: str, hash_type: str, downloaded_hash: str, expected_hash: str):
+        """Initialize."""
+        self.url = url
+        self.hash_type = hash_type
+        self.downloaded_hash = downloaded_hash
+        self.expected_hash = expected_hash
+
+    def __str__(self):
+        """String representation of this exception."""
+        return (
+            f"Failed integrity protection on {self.url!r}, using hash {self.hash_type!r}. "
+            f"Expected: {self.expected_hash!r} but it was {self.downloaded_hash!r}"
+        )
+
+    def __repr__(self):
+        """String representation of this exception."""
+        return self.__str__()
+
+
 class Downloader(Thread):  # pylint:disable=too-many-instance-attributes
     """Log downloader for ETOS client."""
 
@@ -107,10 +129,13 @@ class Downloader(Thread):  # pylint:disable=too-many-instance-attributes
         with open(download_path, "wb+") as report:
             for chunk in response:
                 report.write(chunk)
-        if not self.__verify(item, download_path):
-            self.logger.error("Checksum verification failed on %r", item)
+        try:
+            self.__verify(item, download_path)
+        except IntegrityError:
+            download_path.unlink()
+            raise
 
-    def __verify(self, item: Downloadable, path: Path) -> bool:
+    def __verify(self, item: Downloadable, path: Path):
         """Verify the checksum of the downloaded item.
 
         Path is the real path to the saved file.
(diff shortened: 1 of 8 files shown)


cli/src/etos_client/sse/protocol.py
```diff
@@ -98,6 +98,34 @@ def __str__(self):
         return self.message.get("message")
 
 
+class Report(UserEvent):
```
andmat900, line 101: Is there a need to have one more event containing the list of produced files? 
I. e. an event that is sent at the end of a test run to make sure _all_ reports/artifacts have been downloaded.

cli/src/etos_client/sse/protocol.py
```diff
@@ -98,6 +98,34 @@ def __str__(self):
...
+    """An ETOS test case report file event."""
+
+    def __init__(self, event: dict) -> None:
+        """Initialize a report by loading an expected json string."""
+        super().__init__(event)
+
+        try:
+            self.file = json.loads(self.data)
```
andmat900, line 109: Add checksum? To minimize the risk of files becoming corrupt during upload/download?

t-persson: Added checksum

cli/src/etos_client/sse/protocol.py
t-persson: I don't see the point, if we already get events that say which files to download (the Artifact & Report events) there should be no need for another event saying which files to download.

cli/src/etos_client/downloader/downloader.py
```diff
@@ -142,6 +107,23 @@ def __save_file(self, item: Downloadable, response: requests.Response) -> None:
         with open(download_path, "wb+") as report:
             for chunk in response:
                 report.write(chunk)
+        if not self.__verify(item, download_path):
```
t-persson, line 110: We should probably allow the checksum to be empty and not be verified, for the cases where ETOS is not the producer of the file.

cli/src/etos_client/downloader/downloader.py
```diff
@@ -142,6 +107,49 @@ def __save_file(self, item: Downloadable, response: requests.Response) -> None:
...
+            self.logger.error("Checksum verification failed on %r", item)
+
+    def __verify(self, item: Downloadable, path: Path) -> bool:
+        """Verify the checksum of the downloaded item.
+
+        Path is the real path to the saved file.
+        """
+        translation_table = {
```
andmat900, line 118: sha256 alone would probably be enough

cli/src/etos_client/downloader/downloader.py
```diff
@@ -142,6 +107,49 @@ def __save_file(self, item: Downloadable, response: requests.Response) -> None:
...
+            "SHA-384": "sha384",
+            "SHA-512": "sha512",
+            "SHA-512/224": "sha512_224",
+            "SHA-512/256": "sha512_256"
+        }
+        hash = None
+        expected_digest = None
+        for nist_scheme, python_scheme in translation_table.items():
+            if item.checksums.get(nist_scheme) is not None:
+                hash = hashlib.new(python_scheme)
```
andmat900, line 130: in sse.rst the examples show checksums as empty strings. Should this code also check if the checksum isn't an empty string?

cli/src/etos_client/downloader/downloader.py
```diff
@@ -142,6 +107,49 @@ def __save_file(self, item: Downloadable, response: requests.Response) -> None:
         with open(download_path, "wb+") as report:
             for chunk in response:
                 report.write(chunk)
+        if not self.__verify(item, download_path):
+            self.logger.error("Checksum verification failed on %r", item)
```
andmat900, line 111: Will there only be this log message? Shouldn't it propagate further so that the client can exit with exit code > 0?

cli/src/etos_client/downloader/downloader.py
t-persson: It's enough for ETOS, but with the ETR plugins being able to produce other files for download I think it would be sensible to support all algorithms that are supported by Eiffel, which is the reason I added all of them.

fredjn (approved):

t-persson:
Rebased against main which is the reason for force push. A comparison can be found here: https://github.com/eiffel-community/etos/compare/72ea1bce610228e406717923c83df4cfab5957f3..505b524b9de3fa7021d34f65d2265c0b634df22d

Issue #263, opened 2024-09-18
The status of an environment request should be updated when providers and environment provider are not available


t-persson:
### Description

The environment request controller checks the status of providers but does not update its status if there are any errors.
Similarly the reconcileEnvironmentProvider function may return errors, but we do not update status for that either.

### Further links

https://github.com/eiffel-community/etos/blob/main/internal/controller/environmentrequest_controller.go#L119
https://github.com/eiffel-community/etos/blob/main/internal/controller/environmentrequest_controller.go#L130

PR #319, opened 2024-12-02
Update environment request status on errors

andmat900:
### Applicable Issues
- https://github.com/eiffel-community/etos/issues/263

### Description of the Change

This change adds environment request jobs status updates on errors during provider check and during environment provider reconciliation.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by: andrei.matveyeu@axis.com

### DIFF
--- a/internal/controller/environmentrequest_controller.go
+++ b/internal/controller/environmentrequest_controller.go
@@ -116,19 +116,25 @@ func (r *EnvironmentRequestReconciler) reconcile(ctx context.Context, environmen
 	logger.V(1).Info("environment provider count", "active", len(environmentProviders.activeJobs), "successful", len(environmentProviders.successfulJobs), "failed", len(environmentProviders.failedJobs))
 
 	// Check providers availability
-	// TODO: Update status
 	providers := etosv1alpha1.Providers{
 		IUT:            environmentrequest.Spec.Providers.IUT.ID,
 		ExecutionSpace: environmentrequest.Spec.Providers.ExecutionSpace.ID,
 		LogArea:        environmentrequest.Spec.Providers.LogArea.ID,
 	}
 	if err := checkProviders(ctx, r, environmentrequest.Namespace, providers); err != nil {
+		meta.SetStatusCondition(&environmentrequest.Status.Conditions, metav1.Condition{Status: metav1.ConditionTrue, Type: StatusFailed, Message: "Provider check failed", Reason: err.Error()})
+		if _err := r.Status().Update(ctx, environmentrequest); _err != nil {
+			return _err
+		}
 		return err
 	}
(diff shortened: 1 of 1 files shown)


internal/controller/environmentrequest_controller.go
```diff
@@ -116,19 +116,25 @@ func (r *EnvironmentRequestReconciler) reconcile(ctx context.Context, environmen
...
 
 	// Check providers availability
-	// TODO: Update status
 	providers := etosv1alpha1.Providers{
 		IUT:            environmentrequest.Spec.Providers.IUT.ID,
 		ExecutionSpace: environmentrequest.Spec.Providers.ExecutionSpace.ID,
 		LogArea:        environmentrequest.Spec.Providers.LogArea.ID,
 	}
 	if err := checkProviders(ctx, r, environmentrequest.Namespace, providers); err != nil {
+		meta.SetStatusCondition(&environmentrequest.Status.Conditions, metav1.Condition{Status: metav1.ConditionTrue, Type: StatusFailed, Message: "Provider check failed", Reason: err.Error()})
```
t-persson, line 125: `StatusReady` is the one that we use in the environment request and that is the status that the testrun controller uses to determine if the environmentrequest has completed.

t-persson, line 125: For this to be properly picked up by the testrun controller the reason must be `Failed`.
Reason is a single word state for the status. Example:

NAME            READY   REASON   MESSAGE
test-my-test   False       Failed        Traceback (most recent call last):...

In this case `Ready` is the Status condition on the `StatusReady` type.
Reason is the reason field.
Message is the message field.

t-persson, line 125: Message should be more descriptive

internal/controller/environmentrequest_controller.go
```diff
@@ -116,19 +116,25 @@ func (r *EnvironmentRequestReconciler) reconcile(ctx context.Context, environmen
...
+		if _err := r.Status().Update(ctx, environmentrequest); _err != nil {
```
t-persson, line 126: It's better to log the original error and return `r.Status().Update(...)` here. Since any error returned will look like an error in the controller which this case is not. This is just a reconciliation error.

```
meta.SetStatusCondition(....)
logger.Error(err, "Error getting provider")
return r.Status().Update(...)
```

t-persson (changes requested):

internal/controller/environmentrequest_controller.go
andmat900: Changed to ConditionFalse, StatusReady

t-persson (approved):

PR #325, opened 2024-12-03
Revert "Remove environmentRequest dependency for Environment"

t-persson:
Reverts eiffel-community/etos#324

I didn't think about the bigger picture here. Our environment providers may required a few items from the dataset in particular in order to run their cleanup operations.
The environment request resource must exist for an environment to be cleaned up properly without manual intervention. Having the environment request be required by the environment_controller will enforce this even if it may seem unnecessary to do.

A new PR will be uploaded where we add finalizers to the environmentrequests so that they may not be deleted before the environment has been cleaned up.


internal/controller/environment_controller.go
```diff
@@ -199,17 +199,12 @@ func (r *EnvironmentReconciler) reconcileReleaser(ctx context.Context, releasers
...
-					return err
-				}
-				image = environmentRequest.Spec.Image
+			environmentRequest, err := r.environmentRequest(ctx, environment)
+			if err != nil {
+				return err
 			}
-			releaser := r.releaseJob(environment, *image)
+			releaser := r.releaseJob(environment, environmentRequest)
+			fmt.Println(releaser)
```
andmat900, line 207: Why println? Log?

t-persson: This is a revert.

Issue #268, opened 2024-09-18
We should be able to get termination logs from multiple pods


t-persson:
### Description

When checking for termination logs for jobs we fetch all pods for the job but if there are more than one pod connected to the job we only check the first one.

### Further links

https://github.com/eiffel-community/etos/blob/main/internal/controller/jobs.go#L137

PR #336, opened 2025-01-07
Retrieve termination log from all pods

andmat900:
### Applicable Issues
- https://github.com/eiffel-community/etos/issues/268

### Description of the Change
This enables termination log retrieval from an arbitrary pod of a job.

### Alternate Designs
 The initial proposal includes two alternatives:
- a modified version of the existing `terminationLog()` function which takes pod name prefix as argument
- and a more generic `terminationLogs()` function that returns a map of termination logs for all pods/containers of a job, allowing the caller to process the output

### Sign-off

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/internal/controller/environment_controller.go
+++ b/internal/controller/environment_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -154,16 +155,16 @@ func (r *EnvironmentReconciler) reconcileReleaser(ctx context.Context, releasers
 
 	// Environment releaser failed, setting status.
 	if releasers.failed() {
-		description := ""
+		descriptions := []string{}
 		for _, releaser := range releasers.failedJobs {
 			jobResult, err := terminationLogs(ctx, r, releaser)
 			if err != nil {
-				description = fmt.Sprintf("%s; %s: %s", description, releaser.Name, err.Error())
+				descriptions = append(descriptions, fmt.Sprintf("%s: %s", releaser.Name, err.Error()))
 			} else {
-				description = fmt.Sprintf("%s; %s: %s", description, releaser.Name, jobResult.getConclusion())
+				descriptions = append(descriptions, fmt.Sprintf("%s: %s", releaser.Name, jobResult.getConclusion()))
 			}
-			description = fmt.Sprintf("%s; %s: %s", description, releaser.Name, jobResult.getConclusion())
 		}
+		description := strings.Join(descriptions, ";")
 		if meta.SetStatusCondition(&environment.Status.Conditions, metav1.Condition{Type: StatusActive, Status: metav1.ConditionFalse, Reason: "Failed", Message: description}) {
 			return r.Status().Update(ctx, environment)
 		}
@@ -172,7 +173,7 @@ func (r *EnvironmentReconciler) reconcileReleaser(ctx context.Context, releasers
 	// Environment releaser successful, setting status.
 	if releasers.successful() {
 		reason := ""
-		description := ""
+		descriptions := []string{}
 		for _, releaser := range releasers.successfulJobs {
 			jobResult, err := terminationLogs(ctx, r, releaser)
 			if err != nil {
@@ -180,9 +181,10 @@ func (r *EnvironmentReconciler) reconcileReleaser(ctx context.Context, releasers
 			}
 			if jobResult.getVerdict() == VerdictFailed {
 				reason = "Failed"
-				description = fmt.Sprintf("%s; %s: %s", description, releaser.Name, jobResult.getConclusion())
+				descriptions = append(descriptions, fmt.Sprintf("%s: %s", releaser.Name, jobResult.getConclusion()))
 			}
 		}
+		description := strings.Join(descriptions, ";")
 		if reason == "Failed" {
 			if meta.SetStatusCondition(&environment.Status.Conditions, metav1.Condition{Type: StatusActive, Status: metav1.ConditionFalse, Reason: "Failed", Message: description}) {
 				return r.Status().Update(ctx, environment)
--- a/internal/controller/environment_controller.go
+++ b/internal/controller/environment_controller.go
@@ -154,36 +154,38 @@ func (r *EnvironmentReconciler) reconcileReleaser(ctx context.Context, releasers
 
 	// Environment releaser failed, setting status.
 	if releasers.failed() {
-		releaser := releasers.failedJobs[0] // TODO: We should allow multiple releaser jobs in the future
-		result, err := terminationLog(ctx, r, releaser, releaser.Name, environment.Name)
-		if err != nil {
-			result.Description = err.Error()
+		description := ""
+		for _, releaser := range releasers.failedJobs {
+			jobResult, err := terminationLogs(ctx, r, releaser)
+			if err != nil {
+				description = fmt.Sprintf("%s; %s: %s", description, releaser.Name, err.Error())
+			} else {
+				description = fmt.Sprintf("%s; %s: %s", description, releaser.Name, jobResult.getConclusion())
+			}
+			description = fmt.Sprintf("%s; %s: %s", description, releaser.Name, jobResult.getConclusion())
 		}
-		if result.Description == "" {
-			result.Description = "Failed to release an environment - Unknown error"
-		}
-		if meta.SetStatusCondition(&environment.Status.Conditions, metav1.Condition{Type: StatusActive, Status: metav1.ConditionFalse, Reason: "Failed", Message: result.Description}) {
+		if meta.SetStatusCondition(&environment.Status.Conditions, metav1.Condition{Type: StatusActive, Status: metav1.ConditionFalse, Reason: "Failed", Message: description}) {
 			return r.Status().Update(ctx, environment)
 		}
 	}
 
-	// TODO: L170-199: this is a draft partial implementation of result processing received from terminationLogs() which includes all pods/containers:
 	// Environment releaser successful, setting status.
 	if releasers.successful() {
-		// each releaser is a job that can have multiple pods and each pod can have multiple containers
+		reason := ""
+		description := ""
 		for _, releaser := range releasers.successfulJobs {
-			allPodsResults, err := terminationLogs(ctx, r, releaser)
+			jobResult, err := terminationLogs(ctx, r, releaser)
 			if err != nil {
 				return err
 			}
-			for _, allContainerResults := range allPodsResults {
-				for _, result := range allContainerResults {
-					if result.Conclusion == ConclusionFailed {
-						if meta.SetStatusCondition(&environment.Status.Conditions, metav1.Condition{Type: StatusActive, Status: metav1.ConditionFalse, Reason: "Failed", Message: result.Description}) {
-							return r.Status().Update(ctx, environment)
-						}
-					}
-				}
+			if jobResult.failed() {
+				reason = "Failed"
+				description = fmt.Sprintf("%s; %s: %s", description, releaser.Name, jobResult.getConclusion())
+			}
+		}
+		if reason == "Failed" {
+			if meta.SetStatusCondition(&environment.Status.Conditions, metav1.Condition{Type: StatusActive, Status: metav1.ConditionFalse, Reason: "Failed", Message: description}) {
+				return r.Status().Update(ctx, environment)
 			}
 		}
 		if meta.SetStatusCondition(&environment.Status.Conditions, metav1.Condition{Type: StatusActive, Status: metav1.ConditionFalse, Reason: "Released", Message: "All releaser jobs succeeded"}) {
--- a/internal/controller/environment_controller.go
+++ b/internal/controller/environment_controller.go
@@ -166,19 +166,27 @@ func (r *EnvironmentReconciler) reconcileReleaser(ctx context.Context, releasers
 			return r.Status().Update(ctx, environment)
 		}
 	}
+
+	// TODO: L170-199: this is a draft partial implementation of result processing received from terminationLogs() which includes all pods/containers:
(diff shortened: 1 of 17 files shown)


internal/controller/jobs.go
```diff
@@ -122,8 +123,8 @@ type Result struct {
 	Description string     `json:"description,omitempty"`
 }
 
-// terminationLog reads the termination-log part of the ESR pod and returns it.
-func terminationLog(ctx context.Context, c client.Reader, job *batchv1.Job, containerName string) (*Result, error) {
+// terminationLog reads the termination-log from the container inside the pod given by prefix
+func terminationLog(ctx context.Context, c client.Reader, job *batchv1.Job, podNamePrefix string, containerName string) (*Result, error) {
```
t-persson, line 127: We need to get termination logs from all pods that are relevant, not just a single pod. The reason for this is that the job controller may create more than one pod, for instance when doing retries (which we've disabled) or rolling over to a new node.

Most of the time there will only be a single pod, but we need to check all and this should be the default to always check all.

It is also possible for a single pod having failed and another succeeding (in the case of retries), in that case we don't want to fail because the first pod failed.

andmat900: Then the alternative `terminationLogs()` in L160 is the better option? I. e. the caller gets results from all pods/containers and processes them according to its own logic.

t-persson: Probably, but the callers don't process anything at the moment and need to be updated.

andmat900: This block (L170-199)  is a draft how  this processing could be done by callers:

https://github.com/eiffel-community/etos/pull/336/files#diff-44c6b0e4aeec632561f46bbb19d6e23e953c1df39008d6df513dec474ca3af45

Agreed?

internal/controller/jobs.go
```diff
@@ -133,24 +134,64 @@ func terminationLog(ctx context.Context, c client.Reader, job *batchv1.Job, cont
...
+			}
+		}
+		return &Result{Conclusion: ConclusionFailed}, errors.New("found no container status for pod")
 	}
-	pod := pods.Items[0]
+	return &Result{Conclusion: ConclusionFailed}, errors.New("found no pod with the given name")
+}
+
+// terminationLogs reads termination-log for each pod/container of the given job returning it as a map (keys: pod names, values: Result instances)
+func terminationLogs(ctx context.Context, c client.Reader, job *batchv1.Job) (map[string][]*Result, error) {
```
t-persson, line 160: We should still get termination log from a specific container, since we don't write a result struct in the termination logs for all containers in a pod.

internal/controller/environment_controller.go
```diff
@@ -166,19 +166,27 @@ func (r *EnvironmentReconciler) reconcileReleaser(ctx context.Context, releasers
...
+			for _, allContainerResults := range allPodsResults {
+				for _, result := range allContainerResults {
+					if result.Conclusion == ConclusionFailed {
+						if meta.SetStatusCondition(&environment.Status.Conditions, metav1.Condition{Type: StatusActive, Status: metav1.ConditionFalse, Reason: "Failed", Message: result.Description}) {
+							return r.Status().Update(ctx, environment)
+						}
+					}
+				}
 			}
 		}
```
t-persson, line 188: To simplify this we could add a `Results` struct in `jobs.go` with a method for `failed` etc. This struct can then be returned from `terminationLogs`

```
type Results struct {
  results []Result
}

func (r Results) failed() bool {
  for _, r := range r.results {
    if r.Conclusion == ConclusionFailed {
      return true
    }
  }
  return false
}
```

internal/controller/jobs.go
t-persson, line 160: Instead of returning a `map[string][]*Result` we should return `[]*Result` (weird that we are returning a pointer, but that's legacy) or a `Results` (see other comment)

fredjn (approved):

internal/controller/environment_controller.go
andmat900: A quick draft implementation of this approach (L127-163):  https://github.com/eiffel-community/etos/pull/336/files#diff-061d5aa98b57fa713e1b5c779bebaa2d8d5e93525156aa88bcf381f3e66a6e79

Agreed?

internal/controller/jobs.go
andmat900: Getting termination log from a specific container will be possible via a dedicated method inside  the proposed `JobResult` struct:  (L127-163): https://github.com/eiffel-community/etos/pull/336/files#diff-061d5aa98b57fa713e1b5c779bebaa2d8d5e93525156aa88bcf381f3e66a6e79

internal/controller/jobs.go
andmat900: Updated the controllers to use `terminationLogs()` instead of `terminationLog()`.

internal/controller/jobs.go
```diff
@@ -115,42 +115,135 @@ const (
...
 type Result struct {
+	Name        string     `json:"name"`
 	Conclusion  Conclusion `json:"conclusion"`
 	Verdict     Verdict    `json:"verdict,omitempty"`
 	Description string     `json:"description,omitempty"`
 }
 
-// terminationLog reads the termination-log part of the ESR pod and returns it.
-func terminationLog(ctx context.Context, c client.Reader, job *batchv1.Job, containerName string) (*Result, error) {
+// TODO: L127-163 This is a draft implementation of structs and functions to handle results from multiple pods/containers inside an ETOS job:
```
t-persson, line 126: Remove

internal/controller/jobs.go
```diff
@@ -115,42 +115,135 @@ const (
...
+// JobResults describes the status and result of an ETOS job which consists of one or more pods
+type JobResults struct {
+	PodResults []PodResults
+}
+
+// PodResults describes the status and result of a single pod of an ETOS job which consists of one or more containers
+type PodResults struct {
+	Name             string
+	ContainerResults []Result
+}
```
t-persson, line 136: Why do we need two structs instead of just having a single `Results` struct?

internal/controller/jobs.go
```diff
@@ -115,42 +115,135 @@ const (
...
+			if result.Conclusion != ConclusionSuccessful {
+				return false
+			}
+		}
+	}
+	return true
+}
+
+// getConclusion returns the conclusion of a job based on individual container conclusions
+func (jr JobResults) getConclusion() Conclusion {
```
t-persson, line 163: If a job has two pods where one failed and one successful due to retry then what conclusion would be returned?

internal/controller/testrun_controller.go
```diff
@@ -289,39 +289,128 @@ func (r *TestRunReconciler) reconcileEnvironmentRequest(ctx context.Context, tes
...
+	} else if hasInconclusive {
+		// at least one Inconclusive and no Aborted/TimedOut -> ConclusionAborted
+		return ConclusionInconclusive
+	} else if hasFailed {
+		// at least one Failed and no Aborted/TimedOut/Inconclusive -> ConclusionFailed
+		return ConclusionFailed
+	}
+	// no Aborted/TimedOut/Inconclusive/Failed
+	return ConclusionSuccessful
+}
```
t-persson, line 359: I don't understand the purpose of these

internal/controller/testrun_controller.go
```diff
@@ -289,39 +289,128 @@ func (r *TestRunReconciler) reconcileEnvironmentRequest(ctx context.Context, tes
...
+
+// TODO: move to a separate struct in jobs.go?
```
t-persson, line 361: No?

t-persson (changes requested):

internal/controller/testrun_controller.go
andmat900: Code moved to jobs.go.

internal/controller/testrun_controller.go
andmat900: In the latest commit `Result`, `PodResult`, `JobResult` are refactored to a clearer structure.

internal/controller/jobs.go
andmat900: This code has been removed.

internal/controller/jobs.go
andmat900: We need to be able to determine conclusions and verdicts on different levels: container/pod/job/job group.

internal/controller/jobs.go
andmat900: Done

internal/controller/jobs.go
andmat900: This code has been refactored to return a `JobResults` instance.

internal/controller/jobs.go
t-persson: Why?
Don't we just want to get the termination log from the relevant Job. The callers would then need to do "Hey, I have this job, what's the result of the container with name=<myname>?"
What happens internally in the terminationLogs function should not matter to the caller.
Internally in the terminationLogs we would need to check which is the "current" pod that we should check result from, and then get the result of the proper container in that one.

So the API for callers would be the same:
```
result, err := terminationLog(ctx, r, environmentProvider, environmentrequest.Name)
if err != nil {
  result.Description = err.Error()
}
if result.Description == "" {
  result.Description = "Failed to provision an environment - Unknown error"
}
description = fmt.Sprintf("%s; %s: %s", description, environmentProvider.Name, result.Description)
```

where the terminationLogs is something similar to this:

```
func terminationLog(...) (*Result, error) {
  var pods corev1.PodList
  if err := c.List(ctx, &pods, ...); err != nil {
    ...
    return
  }
  if len(pods.Items) == 0 {
    return &Result{...}
  )

  pod := getLatestPodCreated(pods)
  for _, status := pod.Status.ContainerStatues {
     ...
  }
}
```

or if we want to get results from all pods

```
type Results struct {
  Result  // The result of the last created pod (the one that should be relevant)
  Results []Result
}

func terminationLog(...) (*Results, error) {
  var pods corev1.PodList
  if err := c.List(ctx, &pods, ...); err != nil {
    ...
    return
  }
  if len(pods.Items) == 0 {
    return &Result{...}
  )
  for _, pod := range pods {
    ...
    if isLatest() {
      results.Description = result.Description
      results.Verdict = result.Verdict
      results.Conclusion = result.Conclusion
    }
    results.Results = append(results.Results, result)
    ...
  }
}
```

internal/controller/testrun_controller.go
```diff
@@ -294,34 +294,53 @@ func (r *TestRunReconciler) reconcileSuiteRunner(ctx context.Context, suiteRunne
...
+				return err
+			}
+			containerResult, err := jobResult.getContainerResult(suiteRunner.Name, suiteRunner.Name)
+			if err != nil {
+				description = fmt.Sprintf("%s; %s: %s", description, suiteRunner.Name, err.Error())
+			} else {
+				description = fmt.Sprintf("%s; %s: %s", description, suiteRunner.Name, containerResult.Verdict)
+			}
+			logger.Info("Suite runner result", "name", suiteRunner.Name, "verdict", jobResult.getVerdict(), "conclusion", jobResult.getConclusion(), "message", containerResult.Verdict)
+			description = fmt.Sprintf("%s; %s: %s", description, suiteRunner.Name, containerResult.Verdict)
```
fredjn, line 312: Isn't this line redundant and also overwrites the error case above depending on how many failed jobs there are?

internal/controller/testrun_controller.go
andmat900: Simplified, now using only one struct: `Result`.

internal/controller/jobs.go
andmat900: Using single `Result` struct.

internal/controller/environmentrequest_controller.go
```diff
@@ -168,31 +168,41 @@ func (r *EnvironmentRequestReconciler) reconcile(ctx context.Context, environmen
...
+		for _, provider := range providers.successfulJobs {
+			jobResult, err := terminationLogs(ctx, r, provider)
+			if err != nil {
+				description = fmt.Sprintf("%s; %s: %s", description, provider.Name, err.Error())
+			} else {
+				description = fmt.Sprintf("%s; %s: %s", description, provider.Name, jobResult.getConclusion())
+			}
+			if jobResult.failed() {
+				reason = "Failed"
+				description = fmt.Sprintf("%s; %s: %s", description, provider.Name, jobResult.getConclusion())
```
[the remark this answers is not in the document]
andmat900: Done.

internal/controller/jobs.go
```diff
@@ -108,49 +109,223 @@ const (
...
+				continue
 			}
-			var result Result
-			if err := json.Unmarshal([]byte(status.State.Terminated.Message), &result); err != nil {
-				logger.Error(err, "failed to unmarshal termination log to a result struct")
-				return &Result{Conclusion: ConclusionFailed, Description: status.State.Terminated.Message}, nil
+
+			var result ContainerResult
+			t_msg := status.State.Terminated.Message
+			// The termination message may not be available for some containers such as etos-log-listener, which are not included in the output.
```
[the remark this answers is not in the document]
andmat900: Done.

internal/controller/testrun_controller.go
andmat900: Done.

fredjn (approved):

internal/controller/jobs.go
t-persson: > 1. Why `getLatestPodCreated()`? The original issue requires multiple pods: [We should be able to get termination logs from multiple pods #268](https://github.com/eiffel-community/etos/issues/268)

That means that there may be multiple pods for a job, and we should get the results from the pod that is currently "active". We still only get the results from one pod, but we need to iterate over all pods to find the one whose results we should be interested in.

> 2. In the last example, why is `isLatest()` determining the `Verdict` and `Conclusion` for all pods? Isn't it the combined result of all pods?

As I wrote above: Our jobs will have at-most a single relevant pod, but they may launch multiple pods if we implement retries or they get rolled over to a new node.

t-persson (approved):

PR #339, opened 2025-01-16
Add environment variables and show environment description

t-persson:
### Description of the Change

Add some environment variables for the environment releaser, add permissions and show environment termination description when doing kubectl get environment.
The permissions are added to be used by the next generation environment provider which is more of a pipeline orchestrated by ETOS in order to keep much stricter control over how environments are released in case of errors.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/internal/etos/etos.go
+++ b/internal/etos/etos.go
@@ -516,6 +516,24 @@ func (r *ETOSDeployment) role(name types.NamespacedName, labelName string) *rbac
 					"create", "get", "list", "watch", "delete",
 				},
 			},
+			{
+				APIGroups: []string{"batch"},
+				Resources: []string{
+					"jobs",
+				},
+				Verbs: []string{
+					"create", "get", "list", "watch", "delete",
+				},
+			},
+			{
+				APIGroups: []string{""},
+				Resources: []string{
+					"pods",
+				},
+				Verbs: []string{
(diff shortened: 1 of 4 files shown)


internal/etos/etos.go
```diff
@@ -516,6 +516,24 @@ func (r *ETOSDeployment) role(name types.NamespacedName, labelName string) *rbac
 					"create", "get", "list", "watch", "delete",
 				},
 			},
+			{
+				APIGroups: []string{"batch"},
+				Resources: []string{
+					"jobs",
+				},
+				Verbs: []string{
+					"create", "get", "list", "watch", "delete",
```
andmat900, line 525: Please add these code comments in this role definition:

1. Why are `create` and `delete` permissions only for environments and jobs?
2. What are the new permissions intended for? They weren't required for ETOS v1alpha to work so far.

t-persson: Since the risk of that code comment becoming stale is high, I think it would be better to add it to the commit message.

t-persson: Are you okay with this answer?

andmat900: Yes, but please add the comment to the commit message as you suggested.

PR #348, opened 2025-01-30
Fix docker push permissions issue in generate-manifest.yml

andmat900:
### Description of the Change

This change replaces `docker-push` command from `Makefile` with explicit call to `docker push` in `generate-manifest.yaml`. The `docker-push`  in `Makefile` seems incompatible with `ghcr.io` login step, likely because they run with different shells/environment variables, so that credentials aren't shared.

Also, `docker push` is done only after the PR is merged, not on every push.

Error example:
https://github.com/eiffel-community/etos/actions/runs/13052056614/job/36414378632

The change makes the `docker push` call in the same way as in other ETOS repos such as `etos-api`, e. g.:
https://github.com/eiffel-community/etos-api/blob/380153ddd50aeeb32df2faaf2b700c702de50a9c/.github/workflows/build-push.yml#L175

### Sign-off

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/.github/workflows/generate-manifest.yml
+++ b/.github/workflows/generate-manifest.yml
@@ -36,7 +36,8 @@ jobs:
           [ "$VERSION" == "main" ] && VERSION=$(echo ${{ github.sha }} | cut -c1-8)
           echo IMAGE_ID=$IMAGE_ID
           echo VERSION=$VERSION
-          make docker-build docker-push split-installer IMG=$IMAGE_ID:$VERSION
+          make docker-build split-installer IMG=$IMAGE_ID:$VERSION
+          docker push $IMAGE_ID:$VERSION
           mkdir -p manifests/controller
           cp dist/* manifests/controller/
       - name: Commit changes
--- a/.github/workflows/generate-manifest.yml
+++ b/.github/workflows/generate-manifest.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - 20250130_ghcr_migration_fix # TODO: restore
   workflow_dispatch:
 
 permissions:
@@ -27,7 +28,7 @@ jobs:
           rm -r dist | true
       - name: Generate manifest file
         run: |
-          IMAGE_ID=ghcr.io/eiffel-community/etos-controller
+          IMAGE_ID=ghcr.io/andmat900/etos-controller # TODO: restore
           # Strip git ref prefix from version
           VERSION=$(echo "${{ github.ref }}" | sed -e 's,.*/\(.*\),\1,')
           # Strip "v" prefix from tag name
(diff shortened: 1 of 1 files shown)


t-persson:
> ### Description of the Change
> This change replaces `docker-push` command from `Makefile` with explicit call to `docker push` in `generate-manifest.yaml`. The `docker-push` in `Makefile` seems incompatible with `ghcr.io` login step, likely because they run with different shells/environment variables, so that credentials aren't shared.
> 
If that's the case then it would not have worked before either, right? Since we did the same thing
* Call docker login
* make docker-push

andmat900:
> > ### Description of the Change
> > This change replaces `docker-push` command from `Makefile` with explicit call to `docker push` in `generate-manifest.yaml`. The `docker-push` in `Makefile` seems incompatible with `ghcr.io` login step, likely because they run with different shells/environment variables, so that credentials aren't shared.
> 
> If that's the case then it would not have worked before either, right? Since we did the same thing
> 
>     * Call docker login
> 
>     * make docker-push

We do login differently now. Previously the step was:

```
    - name: Log into registry
      run: echo "${{ secrets.REGISTRYPASSWORD }}" | docker login registry.nordix.org -u ${{ secrets.REGISTRYUSERNAME }} --password-stdin
```

Now it is:

```
      - name: Login to GitHub Container Registry
        uses: docker/login-action@v3
        with:
          registry: ghcr.io
          username: ${{ github.actor }}
          password: ${{ secrets.GITHUB_TOKEN }}
```
This has likely affected `docker push` when it is run with `make`.

t-persson:
> > > ### Description of the Change
> > > This change replaces `docker-push` command from `Makefile` with explicit call to `docker push` in `generate-manifest.yaml`. The `docker-push` in `Makefile` seems incompatible with `ghcr.io` login step, likely because they run with different shells/environment variables, so that credentials aren't shared.
> > 
> > 
> > If that's the case then it would not have worked before either, right? Since we did the same thing
> > ```
> > * Call docker login
> > 
> > * make docker-push
> > ```
> 
> We do login differently now. Previously the step was:
> 
> ```
>     - name: Log into registry
>       run: echo "${{ secrets.REGISTRYPASSWORD }}" | docker login registry.nordix.org -u ${{ secrets.REGISTRYUSERNAME }} --password-stdin
> ```
> 
> Now it is:
> 
> ```
>       - name: Login to GitHub Container Registry
>         uses: docker/login-action@v3
>         with:
>           registry: ghcr.io
>           username: ${{ github.actor }}
>           password: ${{ secrets.GITHUB_TOKEN }}
> ```
> 
> This has likely affected `docker push` when it is run with `make`.

The login action and the previous `run` command above do the same thing and the `docker login` command that they both use writes a file on the system running the actions. If there were problems with this, then we would've seen them before and we would have seen problems locally when using this Makefile as well (which I don't).

If we have a Makefile command for docker push then I want us to use it so that we always push images the same way.

Can you verify that this change actually fixes the problem?

andmat900:
> Can you verify that this change actually fixes the problem?

As we seen before, this type of changes can be verified only if we merge it. This login / push procedure has been verified in the other ETOS repositories.

t-persson:
> > Can you verify that this change actually fixes the problem?
> 
> As we seen before, this type of changes can be verified only if we merge it. This login / push procedure has been verified in the other ETOS repositories.

You can create a workflow on your fork or another repository on your user for testing if there's a difference between the two ways of pushing

andmat900:
> > > Can you verify that this change actually fixes the problem?
> > 
> > 
> > As we seen before, this type of changes can be verified only if we merge it. This login / push procedure has been verified in the other ETOS repositories.
> 
> You can create a workflow on your fork or another repository on your user for testing if there's a difference between the two ways of pushing

A fork will be outside eiffel-community, it will not have the necessary permissions.

t-persson:
> > > > Can you verify that this change actually fixes the problem?
> > > 
> > > 
> > > As we seen before, this type of changes can be verified only if we merge it. This login / push procedure has been verified in the other ETOS repositories.
> > 
> > 
> > You can create a workflow on your fork or another repository on your user for testing if there's a difference between the two ways of pushing
> 
> A fork will be outside eiffel-community, it will not have the necessary permissions.

Or change the location to `andmat900/etos-controller` instead, to verify. Here are two workflows that I just ran on my fork (had to modify the workflow in this PR a bit since it does not work):

Pushing with `docker push` (this PR): https://github.com/t-persson/etos/actions/runs/13113007083/job/36580859944 
Workflow file: https://github.com/t-persson/etos/actions/runs/13113007083/workflow

Pushing with `make docker-push`: https://github.com/t-persson/etos/actions/runs/13113007099
Workflow file: https://github.com/t-persson/etos/actions/runs/13113007099/workflow

PR #413, opened 2025-05-21
Update OpenTelemetry collector variable (#47)

fredjn:
### Description of the Change
As a way to to uniform how we reference the OpenTelemetry Collector,
we are removing the variable OTEL_COLLECTOR_HOST and replacing it with
the standard OTEL_EXPORTER_OTLP_ENDPOINT.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by: Fredrik Fristedt <<fredrik.fristedt@axis.com>>

### DIFF
--- a/internal/etos/suitestarter/suitestarter.go
+++ b/internal/etos/suitestarter/suitestarter.go
@@ -670,8 +670,8 @@ func (r *ETOSSuiteStarterDeployment) suiteRunnerTemplate(templateName types.Name
                 value: /graveyard
               - name: OTEL_CONTEXT
                 value: {otel_context}
-              - name: OTEL_COLLECTOR_HOST
-                value: {otel_collector_host}
+			  - name: OTEL_EXPORTER_OTLP_ENDPOINT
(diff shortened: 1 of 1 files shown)


internal/etos/suitestarter/suitestarter.go
```diff
@@ -670,8 +670,8 @@ func (r *ETOSSuiteStarterDeployment) suiteRunnerTemplate(templateName types.Name
                 value: /graveyard
               - name: OTEL_CONTEXT
                 value: {otel_context}
-              - name: OTEL_COLLECTOR_HOST
```
t-persson, line 673: Should be enough to just remove it though, right? Considering the etos config here: https://github.com/eiffel-community/etos/blob/main/internal%2Fetos%2Fetos.go#L401 should already be passed in

fredjn: That sounds reasonable, i'll verify it tomorrow, if it is in fact redundant i'll remove it altogether.

Issue #431, opened 2025-09-29
etosctl: misleading error message when --identity is empty


andmat900:
### Description

When an empty `--identity` is given to `etosctl` the error message is an error with three tracebacks suggesting that there is a network connection to the cluster (`socket.error`, `urllib3.exceptions.ReadTimeoutError`, `requests.exceptions.ReadTimeout`).

### Logs

```shell

```

### Expected Behavior

There should be a basic validation that `--identity` is an UUID / purl and a clear error message when the validation fails.

fredjn:
We should also handle the case where the event id is missing from the ER.

andmat900:
> We should also handle the case where the event id is missing from the ER.

@fredjn PR #626 fixes the client-side crash: empty `--identity` now fails fast, and API 422 errors are displayed properly instead of causing a pydantic `ValidationError`.

Event IDs missing from the ER - that needs a server-side check in `etos-api`. Probably a separate issue. With #626, any error the API returns will now be shown correctly to the user.

PR #432, opened 2025-09-29
ETOS Client: basic identity validation in start subcommand

andmat900:
### Applicable Issues
- https://github.com/eiffel-community/etos/issues/431

### Description of the Change
This change adds basic validation of the `--identity` value given to `etosctl`.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/cli/src/etos_client/etos/v0/subcommands/start.py
+++ b/cli/src/etos_client/etos/v0/subcommands/start.py
@@ -26,6 +26,7 @@ import warnings
 from etos_client.types.result import Conclusion, Verdict
 from etos_client.etos.v0.etos import Etos
 from etos_client.sse.v1.client import SSEClient
+from etos_client.shared.validators import start_args_validator, CLIArgsValidationError
 
 from etosctl.command import SubCommand
 from etosctl.models import CommandMeta
@@ -87,6 +88,13 @@ class Start(SubCommand):
 
     def run(self, args: dict) -> None:
         """Start an ETOS testrun."""
+        # Validate arguments
+        try:
+            start_args_validator.validate_args(args)
+        except CLIArgsValidationError as e:
+            self.logger.error(str(e))
+            sys.exit(1)
+
         if args["--download-reports"]:
             warnings.warn(
                 "The '-d/--download-reports' parameter is deprecated",
--- a/cli/src/etos_client/etos/v1alpha/subcommands/start.py
+++ b/cli/src/etos_client/etos/v1alpha/subcommands/start.py
@@ -25,6 +25,7 @@ from etos_client.types.result import Conclusion, Verdict
 from etos_client.etos.v1alpha.etos import Etos
 from etos_client.sse.v2alpha.client import SSEClient as SSEV2AlphaClient
 from etos_client.sse.v1.client import SSEClient as SSEV1Client
+from etos_client.shared.validators import start_args_validator, CLIArgsValidationError
 
 from etosctl.command import SubCommand
 from etosctl.models import CommandMeta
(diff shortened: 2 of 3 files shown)


fredjn:
> The read timeout error suggest to me that we arent handling the error from the etos API correctly. The API should return a bad request and a message that the client should just pass to the user.
> 
> I.e. I would rather we update the request logic and, If necessary, fix the API instead of adding this type of validation to the client

But do we really wnat ot allow an empty `--identity` in the first place, how would that ever result in a successful test run?

fredjn:
> > > The read timeout error suggest to me that we arent handling the error from the etos API correctly. The API should return a bad request and a message that the client should just pass to the user.
> > > I.e. I would rather we update the request logic and, If necessary, fix the API instead of adding this type of validation to the client
> > 
> > 
> > But do we really want ot allow an empty `--identity` in the first place, how would that ever result in a successful test run?
> 
> The point I was trying to make was that the API should do the validation. We could have rudimentary checks in the client, but its important that we make the communication with the API work first

In this case I think it's equally important, the whole point of the client is to make it easier to run ETOS successfully, within that framing/context I think adding validation on both sides (client and server) makes sense.

t-persson:
> > > > The read timeout error suggest to me that we arent handling the error from the etos API correctly. The API should return a bad request and a message that the client should just pass to the user.
> > > > I.e. I would rather we update the request logic and, If necessary, fix the API instead of adding this type of validation to the client
> > > 
> > > 
> > > But do we really want ot allow an empty `--identity` in the first place, how would that ever result in a successful test run?
> > 
> > The point I was trying to make was that the API should do the validation. We could have rudimentary checks in the client, but its important that we make the communication with the API work first
> 
> In this case I think it's equally important, the whole point of the client is to make it easier to run ETOS successfully, within that framing/context I think adding validation on both sides (client and server) makes sense.

Im quite sure we have discussed this before and that consensus was to move all validation to the API instead of the client so that it would be easier to update.
I have not found if we have documented this somewhere during an extremely quick and rudimentary search (I dont have much spare time 😉).

t-persson:
> > > > > > > The read timeout error suggest to me that we arent handling the error from the etos API correctly. The API should return a bad request and a message that the client should just pass to the user.
> > > > > > > I.e. I would rather we update the request logic and, If necessary, fix the API instead of adding this type of validation to the client
> > > > > > 
> > > > > > 
> > > > > > But do we really want ot allow an empty `--identity` in the first place, how would that ever result in a successful test run?
> > > > > 
> > > > > 
> > > > > The point I was trying to make was that the API should do the validation. We could have rudimentary checks in the client, but its important that we make the communication with the API work first
> > > > 
> > > > 
> > > > In this case I think it's equally important, the whole point of the client is to make it easier to run ETOS successfully, within that framing/context I think adding validation on both sides (client and server) makes sense.
> > > 
> > > 
> > > Im quite sure we have discussed this before and that consensus was to move all validation to the API instead of the client so that it would be easier to update. I have not found if we have documented this somewhere during an extremely quick and rudimentary search (I dont have much spare time 😉).
> > 
> > Luckly things change, as do opinions ;) . I guess we can hold off in this if you feel adding validations to the client is to much of a change without further discussion, however I can agree that the referenced issue should fix the server-side as well, since we need proper error messages even if the user isn't using etosctl.,
> 
> It makes sense to implement basic validation in both the ETOS client and ETOS API. This provides a fail-fast approach that prevents error cascading and reduces invalid requests to the API server.
> 
> PR for server-side validation: https://github.com/eiffel-community/etos-api/pull/119/files

 I agreee on the fail-fast, but we can fail pretty fast with just the API. The problem Im having with it is that for a v1alpha API we may make breaking changes and the current client would become useless in that case.

For stable APIs it makes more sense to me to have more validation on the client side, however in that case I would love it if we shipped the API schema with the client

andmat900:
Will be reopened if needed. Validation has been added to ETOS API: https://github.com/eiffel-community/etos-api/pull/119

Issue #443, opened 2025-10-29
Robust provider pipeline


t-persson:
### Summary

We need to have a more robust provider pipeline in ETOS v1 where instead of calling several external services (or JSONTas) in a pipeline and hoping that the external providers are good at keeping their state and manage the resources (IUTs, executors and logareas) so that they are properly checked in etc.

### Context

We should create new resources for IUT, ExecutionSpace and LogArea in Kubernetes for traceability.

These resources shall be created by providers of each type, running sequentially in a single Kubernetes Job.

New controllers for IUT, ExecutionSpace and LogArea shall be created. These new providers shall be responsible for releasing the resources should something go wrong before an Environment is created.

When an Environment is created it shall take over control of the IUT, ExecutionSpace and LogArea resources, making sure that they are released on Environment deletion.

The Provider resources shall support adding an Image in addition to Host and JSONTas. ETOS shall provide images for External and JSONTas providers.

PR #457, opened 2025-11-18
Add executionspace resource

t-persson:
### Applicable Issues

depends-on: #455 
#443 

### Description of the Change

Added a new resource called ExecutionSpace. This is to be used with the provider pipeline to store information about provisioned execution spaces.
This change adds the resource and its controller & webhook providing the functionality for the ExecutionSpace but it is not in use by ETOS and is not to affect the current v0 and v1alpha ETOS versions.

### Sign-off

    indicated in the file; or

    in the file; or

    it.


internal/webhook/v1alpha2/executionspace_webhook.go
```diff
@@ -0,0 +1,94 @@
...
+	executionspacelog.Info("Defaulting for ExecutionSpace", "name", executionspace.GetName())
+
+	environmentrequest := &v1alpha1.EnvironmentRequest{}
+	namespacedName := types.NamespacedName{Name: executionspace.Spec.EnvironmentRequest, Namespace: executionspace.Namespace}
+	if err := cli.Get(ctx, namespacedName, environmentrequest); err != nil {
+		executionspacelog.Error(err, "name", executionspace.Name, "namespace", executionspace.Namespace, "environmentRequest", namespacedName.Name,
+			"Failed to get environmentrequest in namespace")
+	}
+
+	executionspace.Labels["etos.eiffel-community.github.io/environment-request"] = environmentrequest.Spec.Name
```
andmat900: before setting labels does executionspace.Labels needs to be initialized if nil?

internal/webhook/v1alpha2/executionspace_webhook.go
```diff
@@ -0,0 +1,94 @@
...
+	if !ok {
+		return fmt.Errorf("expected an ExecutionSpace object but got %T", obj)
+	}
+	executionspacelog.Info("Defaulting for ExecutionSpace", "name", executionspace.GetName())
+
+	environmentrequest := &v1alpha1.EnvironmentRequest{}
+	namespacedName := types.NamespacedName{Name: executionspace.Spec.EnvironmentRequest, Namespace: executionspace.Namespace}
+	if err := cli.Get(ctx, namespacedName, environmentrequest); err != nil {
+		executionspacelog.Error(err, "name", executionspace.Name, "namespace", executionspace.Namespace, "environmentRequest", namespacedName.Name,
+			"Failed to get environmentrequest in namespace")
```
andmat900: return error if environmentrequest does not exist?

andmat900 (dismissed):

internal/release/release.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// ReleaseContainer returns a container specification.
+func ReleaseContainer(name, containerName, namespace, image string, noDelete bool) corev1.Container {
+	args := []string{
+		"-release",
+		fmt.Sprintf("-namespace=%s", namespace),
+		fmt.Sprintf("-name=%s", name),
+	}
+	if noDelete {
```
fredjn: Perhaps add a comment about what it is that we're not deleting? I assume it is the execution-space resource?

fredjn (changes requested):

internal/controller/executionspace_controller.go
```diff
@@ -0,0 +1,308 @@
...
+		} else {
+			condition = metav1.Condition{
+				Type:    status.StatusActive,
+				Status:  metav1.ConditionFalse,
+				Reason:  status.ReasonCompleted,
+				Message: result.Description,
+			}
+		}
+		executionSpaceCondition := meta.FindStatusCondition(*conditions, status.StatusActive)
+		executionSpace.Status.CompletionTime = &executionSpaceCondition.LastTransitionTime
```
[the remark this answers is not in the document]
t-persson: Noticed that the `CompletionTime` would be wrong as well, since it would set it to an earlier condition change.
Updated so that we use metav1.Time instead.

internal/controller/providers.go
```diff
@@ -55,3 +55,18 @@ func checkProvider(ctx context.Context, c client.Reader, name string, namespace
...
+	return &provider, c.Get(ctx, types.NamespacedName{Name: name, Namespace: namespace}, &provider)
+}
+
+// image returns an image that can be used to execute provider.
+func image(provider *etosv1alpha1.Provider) string {
+	var image string
+	if provider.Spec.Image != "" {
+		return provider.Spec.Image
+	}
+	return image
```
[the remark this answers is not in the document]
t-persson: I should probably update all users of the function as well.

fredjn (dismissed):

fredjn (dismissed):

andmat900 (approved):

internal/webhook/v1alpha2/executionspace_webhook.go
SanthiaM: Good catch on the partition overlap. The `session_started_at` range for a chunk does reach back 30 days into a prior chunk's territory due to the session_facts lookback, so adjacent chunks can write to the same `session_started_at` partitions.

However, a session enters a chunk **only** through `sessions_in_window`, which is gated by `first_impression_event_timestamp >= start_at AND < end_at`. Since a session's first-impression timestamp falls in exactly one chunk window, `(session_id, section_id_clean)` merge keys are disjoint between parallel chunks — satisfying er_model_runner Option 1. The physical partition overlap doesn't produce logical key conflicts or concurrent DML races on the same rows.

Added an inline comment to the DAG to document this reasoning for future readers.

PR #459, opened 2025-11-18
Environmentrequest controller refactoring

t-persson:
### Applicable Issues

depends-on: #454 
#443 

### Description of the Change

Exit early if environmentrequest has already failed. Removes a few API calls to the Kubernetes API since we cannot succeed.
Fix a nasty reconciliation loop that could happen if the environment provider jobs fail.
Move out environment deletion to its own function.
Remove the TTL of the environment provider job. This makes sure the job is alive for as long as the environment request is alive, making it easier to debug.
Make sure we handle an empty identifier. There was a TODO in the code.
Register the environmentrequest as the environment owner. This is causes the environmentrequest controller to reconcile if there are updates to the environment.
Added testrun ID to the output when listing environmentrequests with kubectl.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/internal/controller/utilities.go
+++ b/internal/controller/utilities.go
@@ -58,6 +58,8 @@ func hasOwner(ownerReferences []metav1.OwnerReference, kind string) bool {
 }
 
 // isStatusReason return true when the conditionType is present and reason is set to reason
+//
+//nolint:unparam // golangci-lint does not like that all calls of this function pass ReasonFailed
 func isStatusReason(conditions []metav1.Condition, conditionType, reason string) bool {
 	if condition := meta.FindStatusCondition(conditions, conditionType); condition == nil {
 		return false


internal/controller/environmentrequest_controller.go
```diff
@@ -207,7 +214,12 @@ func (r *EnvironmentRequestReconciler) reconcileEnvironmentProvider(ctx context.
 			return nil
 		}
 		if err := jobManager.Create(ctx, environmentrequest, r.environmentProviderJob); err != nil {
-			if meta.SetStatusCondition(conditions,
+			logger.Error(err, "Failed to create environment provider job")
```
andmat900: Does this block have some kind of control how many attempts are made?

Do creation attempts need to be tracked by a counter?

andmat900: Also, would exponential backoff be useful here?

internal/controller/environmentrequest_controller.go
```diff
@@ -466,8 +491,7 @@ func (r EnvironmentRequestReconciler) environmentProviderJob(ctx context.Context
 			Namespace:    environmentrequest.Namespace,
 		},
 		Spec: batchv1.JobSpec{
-			TTLSecondsAfterFinished: &ttl,
-			BackoffLimit:            &backoff,
+			BackoffLimit: &backoff,
```
andmat900: If ttl is removed, will completed/fail jobs persist indefinitely?

internal/controller/environmentrequest_controller.go
t-persson: > Does this block have some kind of control how many attempts are made?

This block, no. It does set the failed reason which stops it from reconciling again and the testrun controller will set Failed as well and shut down deleting the environment request.

> Do creation attempts need to be tracked by a counter?

No

> Also, would exponential backoff be useful here?

The Kubernetes controller-runtime manages this.

fredjn (approved):

PR #486, opened 2025-12-10
Update etos_lib to 5.1.6

andmat900:
This PR updates etos_lib from version 5.1.4 to 5.1.6 in the CLI component to align with other ETOS components.


t-persson:
I'll investigate the e2e-test failure. It is nor related to this change and should hopefully work if you do a retry

t-persson (approved):

Issue #473, opened 2025-11-26
Default service versions when deploying ETOS clusters


t-persson:
### Summary

**As** System owner
**I want** to have the ETOS controller deploy the versions that it is compatible with
**So that** I don't have to manage a cluster spec with ETOS versions locally

### Context

The cluster spec that we are deploying for ETOS holds all service versions, but the defaults are not sane (they point to the :latest tags which don't even exist) meaning I, as a system owner, have to figure out which versions that I should deploy without knowing if the versions are even compatible with eachother.

What I, as a system owner, want is for the ETOS controller to release with versions of each ETOS service that are compatible meaning that I do not ever have to change the defaults in the cluster spec unless I explicitly need to in order to mitigate a certain bug.

As an ETOS maintainer I should still be able to run tests using different versions than the defaults which I can do using the TestRun spec and we should still support overriding versions in the Cluster Spec.

### Acceptance Criteria

- It shall be possible to deploy a Cluster without any versions and get a working ETOS deployment
- It shall be possible to override versions in the Cluster Spec
- It shall be possible to set versions in the Testrun Spec (possible today, so don't break it)
- We shall document all of this so that our users understand the philosophy

PR #505, opened 2026-02-06
Automatically upgrade clusters on new controller versions

t-persson:
### Applicable Issues

fixes: #473 

### Description of the Change

Added defaults to the controller which are embedded into the binary. These files can be updated as before, setting the new image versions. They have, however, moved and changed format so an update to the workflow was necessary.
These defaults will be used if image versions are not set in the cluster specification and will automatically upgrade services if the controller updates (unless set explicitly in the cluster spec).

I also added some e2e tests for this scenario, but I really had to split up the tests a bit.
The e2e_controller_test, e2e_cluster_test and e2e_testrun_test are old tests that I just split out of the primary test file, e2e_defaults_test is the new test scenario.

### Alternate Designs

I investigated whether to use go:embed or go:gen for the default version. With go:gen I could generate a versions struct in the New function of the Config.
I went with the go:embed version since I felt that go:gen would make the code more unreadable and a bit more magic.
We also could have the default version added to the code instead of having separate YAML files, but that would require workflows out in our services to update go source-files and that feels like a dark pattern that I don't want.

### Possible Drawbacks

We are adding "unnecessary" files to the binary, but the alternatives felt a lot worse and either required manual updates of the default versions, update go source-files or generate code when building.
Another drawback is that the Cluster spec does no longer show the full reality of things, since we are deploying Images that are not in the spec.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -18,7 +18,7 @@ package config
 import (
 	"fmt"
 
-	. "github.com/eiffel-community/etos"
+	"github.com/eiffel-community/etos"
 	etosv1alpha1 "github.com/eiffel-community/etos/api/v1alpha1"
 	"gopkg.in/yaml.v3"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


internal/config/config.go
```diff
@@ -0,0 +1,100 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package config
+
+import (
+	"fmt"
+
+	. "github.com/eiffel-community/etos"
```
andmat900: Why not use namespace instead of dot? The dot import makes unclear where Defaults come from.

PR #613, opened 2026-04-13
Add opentelemetry and userlogs to providers

t-persson:
### Applicable Issues

#443 

### Description of the Change

Add opentelemetry spans and userlogs to the provider pipeline.
Also added opentelemetry to the ETOS controller itself. Requires the 'OTEL_EXPORTER_OTLP_ENDPOINT' to be set in order for it to be used. If enabled, the testrun controller will be able to record the 'start-etos' span if it is not set by the ETOS API. Cases where this may be desirable is when creating TestRun resources manually.

I also rewrote the environmentprovider to more closely match the other providers. Mostly so that I would not have to add opentelemetry and logging implementations to that provider.

### Sign-off

    indicated in the file; or

    in the file; or

    it.


pkg/provider/logarea.go
```diff
@@ -34,19 +34,7 @@ import (
...
-	ctx := context.TODO()
-	logger := params.logger.WithValues(
-		"providerType", params.providerType,
-		"environmentRequest", params.environmentRequestName,
-		"namespace", params.namespace,
-		"providerName", params.providerName,
-	)
-	ctx = logr.NewContext(ctx, logger)
-	if err := writeTerminationLog(ctx, runProvider, provider, params); err != nil {
+	if err := run(provider, ParseParameters("ExecutionSpace", GetIUTCount)); err != nil {
```
andmat900: ParseParameters("LogArea" ...

pkg/logging/logging.go
```diff
@@ -205,7 +205,8 @@ func (u *userLogs) Write(p []byte) (n int, err error) {
 	}
 	routingKey := fmt.Sprintf("%s.message.%s", e.Identifier, e.Level)
 	if err := u.publisher.Publish(message, routingKey); err != nil {
-		return 0, fmt.Errorf("failed to publish log entry: %w", err)
+		fmt.Printf("failed to publish log entry: %s\n", err.Error())
```
andmat900, line 215: Invisible errors. Should there be a counter/metric?

internal/controller/testrun_controller.go
```diff
@@ -154,6 +167,69 @@ func (r *TestRunReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	if err != nil {
+		logger.Error(err, "Failed to set baggage member for TestRun ID", "id", testrun.Spec.ID)
+	}
+	b, err = b.SetMember(artifactIDMemberOrDie(testrun.Spec.Artifact))
+	if err != nil {
+		logger.Error(err, "Failed to set baggage member for TestRun ID", "id", testrun.Spec.ID)
+	}
+	b, err = b.SetMember(clusterMemberOrDie(testrun.Spec.Cluster))
+	if err != nil {
+		logger.Error(err, "Failed to set baggage member for TestRun ID", "id", testrun.Spec.ID)
```
andmat900: same error message in three different b.SetMember calls

internal/controller/testrun_controller.go
```diff
@@ -154,6 +167,69 @@ func (r *TestRunReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		attribute.String("etos.artifact.identity", testrun.Spec.Identity),
+	)
+
+	metadata := map[string]string{}
+	otel.GetTextMapPropagator().Inject(ctx, propagation.MapCarrier(metadata))
+	b := baggage.FromContext(ctx)
+
+	b, err = b.SetMember(testRunIDMemberOrDie(testrun))
+	if err != nil {
+		logger.Error(err, "Failed to set baggage member for TestRun ID", "id", testrun.Spec.ID)
```
andmat900, line 209: the *OrDie functions have a panic, the logging on error condition will not execute

internal/controller/testrun_controller.go
```diff
@@ -154,6 +167,69 @@ func (r *TestRunReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	var err error
+	logger.Info("Starting span for TestRun creation", "name", testrun.Name)
+	ctx, span := r.Tracer.Start(
+		ctx, "start-etos", trace.WithSpanKind(trace.SpanKindInternal),
+	)
+	defer func() {
+		err = r.Update(ctx, testrun)
+		if err == nil {
+			// Only end the span if we successfully updated the TestRun with the traceparent annotation.
+			span.End()
```
andmat900, line 193: if the update fails, what will end the span?

internal/controller/testrun_controller.go
```diff
@@ -564,9 +640,9 @@ func (r TestRunReconciler) environmentRequest(ctx context.Context, name string,
 	if ok {
 		annotations["etos.eiffel-community.github.io/traceparent"] = traceparent
 	}
-	baggage, ok := testrun.Annotations["etos.eiffel-community.github.io/baggage"]
+	b, ok := testrun.Annotations["etos.eiffel-community.github.io/baggage"]
```
andmat900, line 647: should tracestate also be propagated?

pkg/provider/provider.go
```diff
@@ -84,23 +98,123 @@ func init() {
 }
 
 // ParseParameters parses the input parameters for a provider.
-func ParseParameters() Parameters {
-	opts := zap.Options{
-		Development: true,
-	}
+func ParseParameters(providerType string, amountFunc AmountFunc) Parameters {
+	opts := zap.Options{}
```
andmat900, line 102: "Developed: true" removed. Intentional?

cmd/iutprovider/main.go
```diff
@@ -33,37 +36,63 @@ func main() {
...
+		span.SetStatus(codes.Error, "invalid minimum amount of IUTs requested")
+		return err
 	}
 	logger.Info("Provisioning a new IUT for EnvironmentRequest",
 		"EnvironmentRequest", environmentRequest.Name,
 		"Namespace", environmentRequest.Namespace,
 		"Amount", cfg.MinimumAmount,
 	)
+	ctx, span = cfg.Tracer.Start(ctx, "CreateIUTs", trace.WithSpanKind(trace.SpanKindInternal))
+	defer span.End()
```
andmat900, line 66: double "defer span.End()" in the same scope.  The variable span is reassigned, the first defer captures the variable by reference and will end the second span, not the first. The first span will never be ended.

andmat900 (changes requested):

internal/controller/testrun_controller.go
t-persson: It is the `SetMember` method we are catching errors for.

internal/controller/testrun_controller.go
t-persson: Nothing will end it, which means it will not be published and then it will be discarded when the context ends.

pkg/provider/provider.go
t-persson: Yes

cmd/environmentprovider/main.go
```diff
@@ -18,162 +18,108 @@ package main
...
 
 	// TODO: Choose strategy
 	split := splitter.NewRoundRobinSplitter().SetSize(maxPossible)
-	for _, test := range request.Spec.Splitter.Tests {
+	for _, test := range environmentRequest.Spec.Splitter.Tests {
 		split.AddTest(test)
 	}
 	tests := split.Split()
 
+	ctx, span = cfg.Tracer.Start(ctx, "CreateEnvironments", trace.WithSpanKind(trace.SpanKindInternal))
```
andmat900, line 92: Reused span variable and double "defer span.End()" call in the Provision() method. Previously in line 43-44.

pkg/logging/logging.go
```diff
@@ -205,7 +206,14 @@ func (u *userLogs) Write(p []byte) (n int, err error) {
...
 	if err := u.publisher.Publish(message, routingKey); err != nil {
-		return 0, fmt.Errorf("failed to publish log entry: %w", err)
+		target := &messaging.PublisherClosedError{}
+		if errors.As(err, &target) {
+			// If the publisher is closed, we can ignore the error and stop trying to publish logs.
+			fmt.Println(err.Error())
+			return 0, nil
+		}
+		fmt.Printf("failed to publish log entry: %s\n", err.Error())
+		return 0, err
```
andmat900, line 216: Returning an error from Write() will cause the logger to treat this writer as broken: user log publishing stops for the rest of the run. A transient RabbitMQ hiccup would permanently kill user logs, while the permanent PublisherClosedError case is handled gracefully. Should this also return 0, nil?

pkg/logging/logging.go
andmat900: This was marked as resolved, but the original concern (observability) not adressed: dropped user log messages are only visible via fmt.Printf to stdout - no metric or counter. Not a blocker, but worth tracking as a follow-up issue.

andmat900 (changes requested):

pkg/logging/logging.go
t-persson: The Publish method will only return an error if the publisher has not been started or if it has actively been closed.
Other transient hiccups shall be handled by the messaging library, not here.

internal/controller/testrun_controller.go
```diff
@@ -155,6 +168,69 @@ func (r *TestRunReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	}
+	testrun.Annotations["etos.eiffel-community.github.io/traceparent"] = metadata["traceparent"]
+	testrun.Annotations["etos.eiffel-community.github.io/tracestate"] = metadata["tracestate"]
+	testrun.Annotations["etos.eiffel-community.github.io/baggage"] = b.String()
+	logger.Info("Added traceparent, tracestate and baggage labels to TestRun",
+		"traceparent", metadata["traceparent"],
+		"tracestate", metadata["tracestate"],
+		"baggage", metadata["baggage"],
+	)
+	return ctx, err
```
andmat900, line 231: This returns errors from the SetMember() calls, but the r.Update() error in the defer above cannot propagate to the caller because err is not a named return value. 

The defer modifies the local variable after the return value has already been captured. If Update fails, the caller receives nil and won't know the annotation wasn't persisted.

internal/controller/testrun_controller.go
```diff
@@ -155,6 +168,69 @@ func (r *TestRunReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	return ctrl.Result{}, nil
 }
 
+// maybeAddTraceparentLabel checks if the traceparent label is already set on the TestRun, and if not, it adds it to
+// enable distributed tracing.
+func (r *TestRunReconciler) maybeAddTraceparentLabel(
+	originalCtx context.Context,
+	testrun *etosv1alpha1.TestRun,
+) (ctx context.Context, err error) {
+	logger := logf.FromContext(ctx)
```
andmat900: logf.FromContext(originalCtx) ?

internal/controller/testrun_controller.go
```diff
@@ -155,6 +168,69 @@ func (r *TestRunReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	if _, ok := testrun.Annotations["etos.eiffel-community.github.io/traceparent"]; ok {
+		return originalCtx, nil
+	}
+	if r.Tracer == nil {
+		logger.Info("No tracer configured, skipping traceparent injection")
+		return originalCtx, nil
+	}
+	// var err error
```
andmat900: remove?

andmat900 (approved):

PR #636, opened 2026-04-28
Add permissions to read secrets

t-persson:
### Applicable Issues

#443 

### Description of the Change

Permissions needed to read secrets from the cluster.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/internal/etos/etos.go
+++ b/internal/etos/etos.go
@@ -691,6 +691,7 @@ func (r *ETOSDeployment) role(name types.NamespacedName, labelName, clusterName 
 				APIGroups: []string{""},
 				Resources: []string{
 					"pods",
+					"secrets",
 				},
 				Verbs: []string{
 					"get", "list", "watch",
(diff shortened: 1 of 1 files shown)


internal/etos/etos.go
```diff
@@ -691,6 +691,7 @@ func (r *ETOSDeployment) role(name types.NamespacedName, labelName, clusterName
 				APIGroups: []string{""},
 				Resources: []string{
 					"pods",
+					"secrets",
```
andmat900: This gives permission to read all secrets in namespace. Would a more targeted approach be possible?

internal/etos/etos.go
```diff
@@ -691,6 +691,7 @@ func (r *ETOSDeployment) role(name types.NamespacedName, labelName, clusterName
...
 				},
 				Verbs: []string{
 					"get", "list", "watch",
```
andmat900, line 696: Is it necessary to have get/list/watch verbs for secrets? Wouldn't a permission to read specific known secrets be enough?

internal/etos/etos.go
t-persson: Since users are able to add any secret into the clusters specification for certain fields this is not really a simple fix to do. If it were just the secrets ETOS creates, then it would be easy.
I would rather we write an issue for this instead of trying to fix it in this pull request.

andmat900 (approved):

PR #637, opened 2026-04-28
Inject test result timeout into provider env vars

andmat900:
### Applicable Issues

https://github.com/eiffel-community/etos/issues/639

Related to https://github.com/eiffel-community/etos-api/pull/156

### Description of the Change

Add `TestResultTimeout` to `ETOSConfig` CRD (default 86400s) and inject it as `ETOS_DEFAULT_TEST_RESULT_TIMEOUT` env var into IUT and ExecutionSpace providers so etcd key TTLs match the configured timeout.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/api/v1alpha1/cluster_types.go
+++ b/api/v1alpha1/cluster_types.go
@@ -296,6 +296,12 @@ type ETOSConfig struct {
 	// +kubebuilder:default="3600"
 	// +required
 	EnvironmentTimeout string `json:"environmentTimeout,omitempty"`
+	// TestResultTimeout describes the maximum time in seconds that test results are expected to
+	// be available. This value is used as the basis for etcd key TTLs in the IUT and execution
+	// space provider services. Defaults to 86400 seconds (24 hours) if not set.
+	// +kubebuilder:default="86400"
+	// +optional
+	TestResultTimeout string `json:"testResultTimeout,omitempty"`
 	// Source describes the source tag added to all Eiffel events sent by ETOS.
 	// +kubebuilder:default="ETOS"
 	// +required
--- a/config/crd/bases/etos.eiffel-community.github.io_clusters.yaml
+++ b/config/crd/bases/etos.eiffel-community.github.io_clusters.yaml
@@ -245,6 +245,13 @@ spec:
                         description: Source describes the source tag added to all
                           Eiffel events sent by ETOS.
                         type: string
+                      testResultTimeout:
+                        default: "86400"
+                        description: |-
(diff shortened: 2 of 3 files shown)


api/v1alpha1/cluster_types.go
```diff
@@ -296,6 +296,12 @@ type ETOSConfig struct {
 	// +kubebuilder:default="3600"
 	// +required
 	EnvironmentTimeout string `json:"environmentTimeout,omitempty"`
+	// TestResultTimeout describes the maximum time in seconds that test results are expected to
+	// be available. This value is used as the basis for etcd key TTLs in the IUT and execution
+	// space provider services. Defaults to 86400 seconds (24 hours) if not set.
+	// +kubebuilder:default="86400"
+	// +optional
+	TestResultTimeout string `json:"testResultTimeout,omitempty"`
```
t-persson, line 304: Where is this timeout specified today?
Seems weird that we don't already have a timeout for it.

andmat900: It's only in the Python etos-library (ETOS_DEFAULT_TEST_RESULT_TIMEOUT env var, default 86400s). The Go providers never had access to it.

t-persson: The weird thing for me here is don't we have a way of configuring that timeout?

andmat900: It is configurable via the ETOS_DEFAULT_TEST_RESULT_TIMEOUT env var (defaulting to 86400s in etos_lib), but it's not exposed as a CRD field - unlike environmentTimeout and testSuiteTimeout, there's no cluster-level knob for it. 

And the Go providers (which write the etcd keys) never read that env var, so the TTLs were never applied on the v1alpha path. This PR adds a TestResultTimeout field to the CRD so it can be configured in one place and injected into the providers that need it.

t-persson: Since v1alpha no longer uses ETCD I don't think we need this particular change (as this was a fix for ETCD TTLs).

I do, however, think we need to make sure the ETOS_DEFAULT_TEST_RESULT_TIMEOUT is configurable in some way. Preferably on the TestRun level instead of the Cluster level. Maybe as a Deadline field like we do on the EnvironmentRequest resources.

t-persson: We might be able to just remove the ETOS_DEFAULT_TEST_RESULT_TIMEOUT entirely if we go the deadline route and just let the testrun_controller manage it

PR #643, opened 2026-05-04
Convert generateName to RFC1123 DNS subdomain format

### LINKED ISSUE #642: Provider packages may create resources with invalid names
### Description

The `CreateIUT`, `CreateLogArea` & `CreateExecutionSpace` functions in the provider package takes the `environmentrequest.Spec.Name` and appends `iut-`, `execution-space-` or `log-area-`
This solution only works if the test suite name is valid according to RFC 1123.

### Additional Context

```
iut-provider: Iut.etos.eiffel-community.github.io "bla_bla-iut" is invalid: [metadata.generateName: Invalid value: "bla_bla-iut-": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

### Logs

```shell

```

### Expected Behavior

We should be able to create resources in the providers.

### Steps To Reproduce

_No response_

### The version of this project/repo, if applicable

_No response_

### The version/edition of the Eiffel Protocol used, if applicable

_No response_

t-persson:
### Applicable Issues

fixes: #642 

### Description of the Change

Convert generated names to be rfc1123 dns subdomain compatible.

### Alternate Designs

I decided against always converting the name and only converting it if it is generated by the provider package.
Providers that pass a name must handle this themselves.

### Possible Drawbacks

~~If the name exceeds 253 characters then we will trim away the `-iut`, `-log-area` and `-execution-space` suffixes. We don't really need to suffix the resources with it since it should already be obvious that an `Iut` resource is an `iut`.~~
Removed suffix

### Sign-off

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/pkg/provider/executionspace.go
+++ b/pkg/provider/executionspace.go
@@ -113,9 +113,9 @@ func CreateExecutionSpace(
 
 	var generateName string
 	if name == "" {
-		// 253 is the maximum length for a name, and we need to reserve one character for the
-		// hyphen added by generateName, so we use 252 here.
-		generateName = fmt.Sprintf("%s-", toRFC1123(environmentrequest.Spec.Name, 252))
+		// 63 is the maximum length for generateName, and we need to reserve one character for the
+		// hyphen added by generateName, so we use 62 here.
+		generateName = fmt.Sprintf("%s-", toRFC1123(environmentrequest.Spec.Name, 62))
 	}
 
 	isController := false
--- a/pkg/provider/executionspace.go
+++ b/pkg/provider/executionspace.go
@@ -18,7 +18,6 @@ package provider
 import (
 	"context"
 	"fmt"
-	"strings"
 
 	"github.com/eiffel-community/etos/api/v1alpha1"
 	"github.com/eiffel-community/etos/api/v1alpha2"
@@ -114,7 +113,9 @@ func CreateExecutionSpace(
 
 	var generateName string
 	if name == "" {
-		generateName = toRFC1123(fmt.Sprintf("%s-execution-space-", strings.ToLower(environmentrequest.Spec.Name)))
+		// 253 is the maximum length for a name, and we need to reserve one character for the
+		// hyphen added by generateName, so we use 252 here.
+		generateName = fmt.Sprintf("%s-", toRFC1123(environmentrequest.Spec.Name, 252))
 	}
 
 	isController := false
--- a/pkg/provider/executionspace.go
+++ b/pkg/provider/executionspace.go
@@ -114,7 +114,7 @@ func CreateExecutionSpace(
 
 	var generateName string
 	if name == "" {
-		generateName = fmt.Sprintf("%s-execution-space-", strings.ToLower(environmentrequest.Spec.Name))
+		generateName = toRFC1123(fmt.Sprintf("%s-execution-space-", strings.ToLower(environmentrequest.Spec.Name)))
 	}
 
 	isController := false
--- a/pkg/provider/provider.go
+++ b/pkg/provider/provider.go
@@ -261,8 +261,9 @@ func writeTerminationLog(
 	return nil
 }
 
-// toRFC1123 converts a string to a valid RFC1123 dns subdomain by replacing invalid characters
-// with hyphens, converting to lowercase and ensuring it doesn't exceed maxlen characters.
+// toRFC1123 converts a string to a valid RFC1123 dns subdomain (but with a variable length)
+// by replacing invalid characters with hyphens, converting to lowercase and ensuring it
+// doesn't exceed maxlen characters.
 func toRFC1123(s string, maxlen int) string {
 	s = strings.ToLower(s)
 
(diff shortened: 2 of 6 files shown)


pkg/provider/provider.go
```diff
@@ -256,3 +258,25 @@ func writeTerminationLog(
...
 	return nil
 }
+
+// toRFC1123 converts a string to a valid RFC1123 dns subdomain by replacing invalid characters
+// with hyphens, converting to lowercase and ensuring it doesn't exceed 253 characters.
+func toRFC1123(s string) string {
+	s = strings.ToLower(s)
+
+	// Replace any non-alphanumeric character with a hyphen
+	reg := regexp.MustCompile(`[^a-z0-9]+`)
```
andmat900: Can be compiled once on package level instead of doing it in every call.

pkg/provider/executionspace.go
```diff
@@ -114,7 +114,7 @@ func CreateExecutionSpace(
 
 	var generateName string
 	if name == "" {
-		generateName = fmt.Sprintf("%s-execution-space-", strings.ToLower(environmentrequest.Spec.Name))
+		generateName = toRFC1123(fmt.Sprintf("%s-execution-space-", strings.ToLower(environmentrequest.Spec.Name)))
```
andmat900: double ToLower (included in toRFC1123())

pkg/provider/provider.go
```diff
@@ -256,3 +258,25 @@ func writeTerminationLog(
...
+
+	s = strings.Trim(s, "-")
+
+	// 253 characters is the maximum length for an RFC1123 resource name in Kubernetes
+	if len(s) > 253 {
+		s = s[:253]
+	}
+
+	// Ensure it doesn't end with a hyphen after truncation
+	s = strings.TrimSuffix(s, "-")
```
andmat900, line 279: Isn't trailing hyphen needed? Otherwise K8S-generated names will be like `myname-iutabc123` instead of `myname-iut-abc123`?

pkg/provider/provider.go
```diff
@@ -256,3 +260,23 @@ func writeTerminationLog(
...
+// toRFC1123 converts a string to a valid RFC1123 dns subdomain by replacing invalid characters
+// with hyphens, converting to lowercase and ensuring it doesn't exceed maxlen characters.
+func toRFC1123(s string, maxlen int) string {
+	s = strings.ToLower(s)
+
+	// Replace any non-alphanumeric character with a hyphen
+	s = nonAlphanumerics.ReplaceAllString(s, "-")
+	s = strings.Trim(s, "-")
+
+	// 253 characters is the maximum length for an RFC1123 resource name in Kubernetes
```
andmat900: Kubernetes API server seems to add 5 characters suffix, so the final name will be up to 258. Will it work?

pkg/provider/provider.go
```diff
@@ -256,3 +260,23 @@ func writeTerminationLog(
...
+	if len(s) > maxlen {
+		s = s[:maxlen]
+	}
+
+	// Ensure it doesn't end with a hyphen after truncation
+	s = strings.TrimSuffix(s, "-")
```
andmat900, line 279: nit: TrimSuffix() removes only a single trailing hyphen. TrimRight() would remove multiple.

t-persson: strings.Trim removes all duplicate hyphens. strings.TrimSuffix is there just to make sure we don't end with a hyphen after truncating the string.

pkg/provider/provider.go
t-persson: generateName will truncate.
Changed to 63 instead of 253 since generateName always truncates down to that size instead of the larger size.

pkg/provider/provider.go
andmat900: strings.Trim handles leading/trailing hyphens before truncation, agreed. 

Consider: input "abc!!!!def" -> after regex becomes "abc----def" -> after Trim stays "abc----def" -> truncated at 6 > "abc---" -> TrimSuffix produces "abc--". TrimRight(s, "-") would give "abc".

Trivial to fix.

pkg/provider/provider.go
andmat900: If generateName truncates internally then 62 is fine. The doc comment says "dns subdomain" but should say "dns label" since the limit is 63, not 253.

pkg/provider/provider.go
t-persson: We have tests that test this scenario.

https://go.dev/play/p/pjobkX1EoO7

pkg/provider/provider.go
t-persson: Updated docs to clarify that the length is variable.

andmat900 (approved):

PR #663, opened 2026-05-11
Add resource requests and limits to the provider spec

### LINKED ISSUE #631: Add resource requests and limits to the provider spec
### Description

With the provider pipelines we run the initContainers without any requests and limits because we cannot properly guess the correct values. We should add a field in the provider spec which allows users to set requests and limits since certain environments require these to be set.

Also update the controllers so that these values are set on the initContainers.

### Additional Context

_No response_

### Further links

_No response_

andmat900:
### Applicable Issues

Fixes https://github.com/eiffel-community/etos/issues/631

### Description of the Change

Add an optional `resources` field to `ProviderSpec` so users can set resource requests and limits on provider init containers in the environment-provider Job.

### Sign-off

    indicated in the file; or

    in the file; or

    it.


t-persson:
Dont forget the releasers

internal/release/release.go
```diff
@@ -86,14 +86,9 @@ func ReleaseContainer(name, containerName, namespace string, provider *v1alpha1.
...
-	return corev1.Container{
-		Name:            containerName,
-		Image:           provider.Spec.Image,
-		ImagePullPolicy: corev1.PullIfNotPresent,
-		Env:             provider.Spec.Env,
-		EnvFrom:         provider.Spec.EnvFrom,
-		// TODO: Verify these resourceclaims
-		Resources: corev1.ResourceRequirements{
+	resources := provider.Spec.Resources
+	if len(resources.Limits) == 0 && len(resources.Requests) == 0 {
```
t-persson: Let's just remove the defaults we added.
Maybe add them to the provider type instead

t-persson (changes requested):

internal/release/release.go
andmat900: Done

api/v1alpha1/provider_types.go
```diff
@@ -38,6 +39,11 @@ type ProviderSpec struct {
 	// +optional
 	EnvFrom []corev1.EnvFromSource `json:"envFrom,omitempty"`
 
+	// Resources describes compute resource requirements for the provider container.
+	// +kubebuilder:default={}
+	// +optional
+	Resources ProviderResources `json:"resources,omitempty"`
```
t-persson: It is possible to set the defaults in `+kubebuilder:default={}` instead of rolling our own resources struct.

t-persson (changes requested):
.

api/v1alpha1/provider_types.go
```diff
@@ -38,6 +38,11 @@ type ProviderSpec struct {
 	// +optional
 	EnvFrom []corev1.EnvFromSource `json:"envFrom,omitempty"`
 
+	// Resources describes compute resource requirements for the provider container.
+	// +kubebuilder:default={limits: {cpu: "250m", memory: "256Mi"}, requests: {cpu: "250m", memory: "256Mi"}}
```
t-persson: Previously the default requests were lower than what we have here now. Is this change intended?

andmat900: Done

internal/controller/environmentrequest_controller.go
```diff
@@ -749,16 +751,6 @@ func (r EnvironmentRequestReconciler) environmentProviderJob(ctx context.Context
...
-							Resources: corev1.ResourceRequirements{
-								Limits: corev1.ResourceList{
-									corev1.ResourceMemory: resource.MustParse("256Mi"),
-									corev1.ResourceCPU:    resource.MustParse("250m"),
-								},
-								Requests: corev1.ResourceList{
-									corev1.ResourceMemory: resource.MustParse("128Mi"),
-									corev1.ResourceCPU:    resource.MustParse("100m"),
-								},
-							},
```
t-persson, line 761: Sorry for missing this one before.

We should keep resource limits for the environment provider, since it is not a provider_type.

t-persson (changes requested):

internal/controller/environmentrequest_controller.go
andmat900: Done

t-persson (approved):

PR #681, opened 2026-05-25
Clean up defaults handling for clusters

t-persson:
### Applicable Issues

Related to my work with #382 

### Description of the Change

This fixes a small problem where some values were set differently depending on the controller which caused some difficulties when working with providers in the provider pipeline.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/internal/controller/testrun_controller.go
+++ b/internal/controller/testrun_controller.go
@@ -703,50 +703,6 @@ func (r TestRunReconciler) createEnvironmentRequests(ctx context.Context, suite 
 // environmentRequest is the definition for an environment request.
 func (r TestRunReconciler) environmentRequest(ctx context.Context, name, testrunner string, cluster *etosv1alpha1.Cluster, testrun *etosv1alpha1.TestRun, tests []etosv1alpha1.Test, dataset *apiextensionsv1.JSON) *etosv1alpha1.EnvironmentRequest {
 	logger := logf.FromContext(ctx)
-	eventRepository := cluster.Spec.EventRepository.Host
-	if cluster.Spec.ETOS.Config.ETOSEventRepositoryURL != "" {
-		logger.Info("Cluster configured with an event repository URL, using that instead")
-		eventRepository = cluster.Spec.ETOS.Config.ETOSEventRepositoryURL
-	}
-	if eventRepository == "" {
-		// TODO: We must fix a global config to store these default values.
-		eventRepository = fmt.Sprintf("http://%s-graphql:%d/graphql", cluster.Name, 5000)
-	}
-	logger.Info("Event repository configured", "url", eventRepository)
-	etosAPI := cluster.Spec.ETOS.Config.ETOSApiURL
-	if etosAPI == "" {
-		etosAPI = fmt.Sprintf("http://%s-etos-api/api", cluster.Name)
-	}
-	logger.Info("ETOS API configured", "url", etosAPI)
-
-	eiffelMessageBus := cluster.Spec.MessageBus.EiffelMessageBus
-	if cluster.Spec.MessageBus.EiffelMessageBus.Deploy {
-		logger.Info("Cluster deployed with an Eiffel messagebus, using that instead")
-		eiffelMessageBus.Host = fmt.Sprintf("%s-%s", cluster.Name, "rabbitmq")
-	}
-	logger.Info("Eiffel messagebus configured", "url", eiffelMessageBus)
-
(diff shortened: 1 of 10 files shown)


internal/webhook/v1alpha1/cluster_webhook.go
```diff
@@ -63,6 +63,35 @@ func (d *ClusterCustomDefaulter) Default(_ context.Context, obj *etosv1alpha1.Cl
...
+			Deploy:          true,
+			EiffelQueueName: obj.GetName(),
+			Host:            fmt.Sprintf("http://%s-graphql.%s.svc.cluster.local:5000/graphql", obj.GetName(), obj.GetNamespace()),
+			Database: etosv1alpha1.MongoDB{
+				Deploy: true,
+			},
+		}
+	}
+	if obj.Spec.ETOS.Ingress.Enabled {
+		obj.Spec.ETOS.Config.ETOSApiURL = obj.Spec.ETOS.Ingress.Host
```
andmat900: Ingress host is a bare hostname. It needs to be converted to URL here.

andmat900: The new code always overwrites ETOSApiURL when ingress is enabled, even  if the user explicitly provided one.

internal/controller/testrun_controller.go
```diff
@@ -812,14 +768,14 @@ func (r TestRunReconciler) environmentRequest(ctx context.Context, name, testrun
...
-				EtosApi:                             etosAPI,
+				EiffelMessageBus:                    *cluster.Spec.MessageBus.EiffelMessageBus,
+				EtosMessageBus:                      *cluster.Spec.MessageBus.ETOSMessageBus,
+				EtosApi:                             cluster.Spec.ETOS.Config.ETOSApiURL,
 				EncryptionKey:                       cluster.Spec.ETOS.Config.EncryptionKey,
 				RoutingKeyTag:                       cluster.Spec.ETOS.Config.RoutingKeyTag,
-				GraphQlServer:                       eventRepository,
-				EtcdHost:                            databaseHost,
-				EtcdPort:                            databasePort,
+				GraphQlServer:                       cluster.Spec.EventRepository.Host,
```
andmat900, line 776: Since EventRepository and Database are now pointers, these will panic with a nil pointer dereference if the webhook didn't run - e.g., existing Cluster resources created before the webhook was deployed, or if the webhook is temporarily unavailable.

internal/etos/etos.go
```diff
@@ -506,26 +506,13 @@ func (r *ETOSDeployment) ingress(name types.NamespacedName, clusterName string)
...
-	etosApi := fmt.Sprintf("http://%s/api", etosHost)
-	if r.Config.ETOSApiURL != "" {
-		etosApi = r.Config.ETOSApiURL
-	}
-	eventRepository := cluster.Spec.EventRepository.Host
-	if r.Config.ETOSEventRepositoryURL != "" {
-		eventRepository = r.Config.ETOSEventRepositoryURL
-	}
-
 	data := map[string][]byte{
```
andmat900, line 509: Config.ETOSEventRepositoryURL no longer used?

andmat900 (changes requested):

internal/webhook/v1alpha1/cluster_webhook.go
t-persson: Correct. This is by design.

internal/controller/testrun_controller.go
t-persson: This would only really be a problem the first time this particular change is deployed and the value that you are pointing to is already set by default in the previous version.
If the webhook is temporarily unavailable then no new resources can be created at that point.

I am okay with a panic in the, unlikely, scenario described

internal/etos/etos.go
t-persson: Correct. Removed from cluster spec.

internal/controller/testrun_controller.go
andmat900: There's an inconsistency: `cluster_controller.go` properly nil-guards `EventRepository`, while `testrun_controller.go` and `etos.go` do not have these guards. If the position is "the webhook always runs", then the guard in `cluster_controller.go` is unnecessary. If it's "be defensive", then the other two files need guards too. The code should be consistent one way or the other.

andmat900 (approved):

PR #687, opened 2026-05-27
Fix a problem with the connection hash

t-persson:
### Description of the Change

This fix makes sure we don't create a new RabbitMQ connection on each reconcile of a testrun.
The hash would only be unique when using the same struct, not when using different structs (even though they had the same parameters). This fix uses an external package to do the hashing properly instead of us implementing our own.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/internal/messaging/pool.go
+++ b/internal/messaging/pool.go
@@ -17,7 +17,6 @@ package messaging
 
 import (
 	"context"
-	"crypto/sha256"
 	"errors"
 	"fmt"
 	"sync"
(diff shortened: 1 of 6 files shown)


internal/messaging/pool.go
```diff
@@ -47,10 +60,12 @@ func (p *PublisherPool) GetPublisher(
 	p.lock.Lock()
 	defer p.lock.Unlock()
 	// Hash the connection parameters and namespace to create a unique name for the connection
-	var hashInput []byte
-	hashInput = fmt.Appendf(hashInput, "%v", parameters)
-	hashInput = fmt.Append(hashInput, namespace)
-	name := fmt.Sprintf("%x", sha256.Sum256(hashInput))
+	namespacedParams := namespacedParameters{parameters, namespace}
+	hash, err := hashstructure.Hash(namespacedParams, nil)
```
andmat900, line 64: Glad to see the hashing is no longer fragile :)  https://github.com/eiffel-community/etos/pull/668#discussion_r3234515745

andmat900 (approved):

internal/messaging/pool.go
t-persson: You never answered my follow-up question and then you approved the PR.

andmat900: Your question was rhetorical, essentially saying "that makes no sense".

t-persson: No, it was not

PR #686, opened 2026-05-26
Move base environment variables to executionspace creation

t-persson:
### Applicable Issues

There's no issue for this fix. I just caught this when I wanted to add new environment variables to our default set and I thought it was weird that all our users need to update their custom execution space providers if they have any.

### Description of the Change

This will make it easier to create executionspace providers, where the custom provider will just add environment variables it needs added or want to override instead of knowing all variables that are required for ETOS.

### Sign-off

Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.

Signed-off-by: Tobias Persson <tobias.persson@axis.com>

PR #696, opened 2026-06-01
Separate generating a resource and creating it

t-persson:
### Applicable Issues

#686 

### Description of the Change

When moving environment variable generation into the provider package we created a small problem where, for example, an ETR job needs to be created before the ExecutionSpace is created in Kubernetes. Since we would not have the environment variables available the ETR could not start properly.
Breaking out generation and creation gives the possibility to first generate the ExecutionSpace, launch ETR and then create the ExecutionSpace.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/cmd/executionspaceprovider/main.go
+++ b/cmd/executionspaceprovider/main.go
@@ -114,7 +114,7 @@ func (p *genericExecutionSpaceProvider) createExecutionSpaces(
 		)
 		environment["ENVIRONMENT_ID"] = id
 		environment["ENVIRONMENT_URL"] = fmt.Sprintf("%s/v1alpha/testrun/%s", cfg.EnvironmentRequest.Spec.Config.EtosApi, id)
-		executionSpace, err := provider.CreateExecutionSpace(ctx, cfg.EnvironmentRequest, cfg.Namespace, "",
+		executionSpace, err := provider.NewExecutionSpace(ctx, cfg.EnvironmentRequest, cfg.Namespace, "",
 			v1alpha2.ExecutionSpaceSpec{
 				ID:         id,
 				TestRunner: testrunner,
@@ -127,12 +127,17 @@ func (p *genericExecutionSpaceProvider) createExecutionSpaces(
 			})
 		if err != nil {
 			span.RecordError(err)
-			span.SetStatus(codes.Error, "failed to create ExecutionSpace")
+			span.SetStatus(codes.Error, "failed to create a ExecutionSpace")
+			return err
+		}
+		if err = executionSpace.Create(ctx); err != nil {
+			span.RecordError(err)
+			span.SetStatus(codes.Error, "call to Kubernetes API to create ExecutionSpace failed")
 			return err
 		}
 		logger.Info(fmt.Sprintf("ExecutionSpace created with name '%s', launching ETOS test runner",
 			executionSpace.Name))
-		if err := p.start(ctx, cfg.EnvironmentRequest, executionSpace); err != nil {
+		if err := p.start(ctx, cfg.EnvironmentRequest, executionSpace.ExecutionSpace); err != nil {
 			span.RecordError(err)
 			span.SetStatus(codes.Error, "failed to start ETOS test runner")
 			return err
--- a/cmd/executionspaceprovider/main.go
+++ b/cmd/executionspaceprovider/main.go
@@ -127,7 +127,7 @@ func (p *genericExecutionSpaceProvider) createExecutionSpaces(
 			})
 		if err != nil {
 			span.RecordError(err)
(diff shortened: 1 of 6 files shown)


pkg/provider/executionspace.go
```diff
@@ -74,27 +73,26 @@ func GetExecutionSpaces(
...
+
+// NewExecutionSpace creates a new ExecutionSpace resource.
 //
 // The spec.ProviderID and spec.EnvironmentRequest fields are automatically populated by this
 // function. It will be overwritten if set.
 // If a name is not provided, a name will be generated based on the EnvironmentRequest name.
 // If a name is provided it is the caller's responsibility to ensure name uniqueness, it will
 // not be guaranteed by this function.
-func CreateExecutionSpace(
+// func CreateExecutionSpace(
```
andmat900: leftover commented-out code

pkg/provider/executionspace.go
```diff
@@ -74,27 +73,26 @@ func GetExecutionSpaces(
 	return executionSpaces, err
 }
 
-// CreateExecutionSpace creates a new ExecutionSpace resource in Kubernetes.
+type ExecutionSpace struct {
+	*v1alpha2.ExecutionSpace
+}
+
+// NewExecutionSpace creates a new ExecutionSpace resource.
```
andmat900: The doc comment on each New* function still reads "creates a new ExecutionSpace resource." Since the whole point of the PR is to distinguish generating from creating in Kubernetes, the comment is  misleading.  Could be something like " "Builds a new ExecutionSpace object in memory; call Create to persist it to Kubernetes."

cmd/iutprovider/main.go
```diff
@@ -67,12 +67,17 @@ func (p *genericIutProvider) createIUTs(ctx context.Context, cfg provider.Provis
 
 	for range cfg.MinimumAmount {
 		logger.Info("Creating a generic IUT")
-		iut, err := provider.CreateIUT(ctx, cfg.EnvironmentRequest, cfg.Namespace, "", v1alpha2.IutSpec{})
+		iut, err := provider.NewIUT(ctx, cfg.EnvironmentRequest, cfg.Namespace, "", v1alpha2.IutSpec{})
 		if err != nil {
 			span.RecordError(err)
 			span.SetStatus(codes.Error, "failed to create IUT")
```
andmat900, line 73: When NewIUT fails, the span still reports "failed to create IUT". It's now a generation failure, not a creation failure.

andmat900 (changes requested):

pkg/provider/executionspace.go
t-persson: This is a draft. Not ready for review.

pkg/provider/executionspace.go
t-persson: This is a draft. Not ready for review.

cmd/iutprovider/main.go
t-persson: This is a draft. Not ready for review.

cmd/executionspaceprovider/main.go
```diff
@@ -127,12 +127,17 @@ func (p *genericExecutionSpaceProvider) createExecutionSpaces(
 			})
 		if err != nil {
 			span.RecordError(err)
-			span.SetStatus(codes.Error, "failed to create ExecutionSpace")
+			span.SetStatus(codes.Error, "failed to create a ExecutionSpace")
```
andmat900: "an ExecutionSpace"

andmat900: The span message should not be saying "create", since the creation is the next step. Should be rather "instantiate" / "generate" or similar.

cmd/iutprovider/main.go
andmat900: still applicable, see the executionspace comment above

cmd/logareaprovider/main.go
```diff
@@ -104,6 +104,11 @@ func (p *genericLogAreaProvider) createLogAreas(
 			span.SetStatus(codes.Error, "failed to create LogArea")
```
andmat900, line 104: same as above

cmd/executionspaceprovider/main.go
t-persson: "generate" and "instantiate" are not correct either since we are creating an ExecutionSpace struct.
I instead clarified the message below when we are creating the executionspace resource in Kubernetes.

t-persson:
```suggestion
			span.SetStatus(codes.Error, "failed to create an ExecutionSpace")
```

andmat900 (approved):

PR #704, opened 2026-06-05
Add event parsing and add instance to status

t-persson:
### Applicable Issues

#382 

### Description of the Change

This will help when we add an SSE client for the providers
Also added an EventType type
Added a Ping event

### Sign-off

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/pkg/messaging/events/events.go
+++ b/pkg/messaging/events/events.go
@@ -15,8 +15,62 @@
 // limitations under the License.
 package events
 
+import (
+	"encoding/json"
+	"fmt"
+)
+
 type Event interface {
-	EventType() string
+	EventType() EventType
 	EventData() any
 	EventMeta() string
+}
+
+type BareEvent struct {
+	ID    int       `json:"id,omitempty"`
+	Event EventType `json:"event"`
+	Data  any       `json:"data"`
+}
+
+type ValidationError struct {
+	Message string
+}
+
+// Error returns the error message of the ValidationError.
+func (e ValidationError) Error() string {
+	return e.Message
+}
+
+// Parse takes a BareEvent and attempts to parse it into a specific Event type based on the EventType.
+func Parse(event BareEvent) (Event, error) {
+	switch event.Event {
+	case MessageType:
+		return parseEvent[Message](event)
+	case ArtifactType:
+		return parseEvent[Artifact](event)
+	case ReportType:
+		return parseEvent[Report](event)
+	case ShutdownType:
+		return parseEvent[Shutdown](event)
+	case StatusType:
+		return parseEvent[Status](event)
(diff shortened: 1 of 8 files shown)


pkg/messaging/events/events.go
```diff
@@ -15,8 +15,62 @@
...
-	EventType() string
+	EventType() EventType
 	EventData() any
 	EventMeta() string
 }
+
+type BareEvent struct {
+	ID    int       `json:"id,omitempty"`
+	Event EventType `json:"event"`
+	Data  any       `json:"data"`
```
andmat900, line 32: no Meta?

pkg/messaging/events/types.go
```diff
@@ -15,6 +15,29 @@
 // limitations under the License.
 package events
 
+import "strings"
+
+type EventType string
+
+var (
+	MessageType  EventType = "message"
```
andmat900, line 23: these should be const, not var

pkg/messaging/events/events.go
```diff
@@ -15,8 +15,62 @@
...
+		return parseEvent[Ping](event)
+	default:
+		return nil, &ValidationError{Message: fmt.Sprintf("unsupported event type: %s", event.Event)}
+	}
+}
+
+// parseEvent is a generic function that takes a BareEvent and attempts to unmarshal it into the specified Event type T.
+func parseEvent[T Event](event BareEvent) (T, error) {
+	var result T
+	jsonData, err := json.Marshal(event)
```
andmat900, line 67: The caller must already have unmarshaled raw JSON into BareEvent; Parse then marshals it back to JSON and unmarshals again

pkg/messaging/events/types.go
```diff
@@ -15,6 +15,29 @@
...
+	StatusType   EventType = "status"
+	PingType     EventType = "ping"
+)
+
+// String returns the string representation of the EventType.
+func (e EventType) String() string {
+	return string(e)
+}
+
+// ToLower returns the lowercase string representation of the EventType.
```
andmat900, line 36: event types are all lowercase already

pkg/messaging/events/events.go
t-persson: Meta serves no purpose when reading, only when publishing and for that we shall use the real event structs.

pkg/messaging/events/events.go
t-persson: That is correct. We convert from BareEvent to T

pkg/messaging/events/types.go
t-persson: Not necessarily when parsing

andmat900 (approved):

PR #705, opened 2026-06-05
Execution space provider can now wait for ETR to start

t-persson:
### Description of the Change

Added an SSE client for subscribing to the SSEv2 stream and set up so that the generic execution space provider now waits for the ETR to report via status events that it is up and running. The wait for test runner exported function was also added for external providers.

### Sign-off

    indicated in the file; or

    in the file; or

    it.


pkg/provider/executionspace.go
```diff
@@ -151,6 +154,45 @@ func (e *ExecutionSpace) Create(ctx context.Context) error {
...
+) error {
+	logger := logging.FromContextOrDiscard(ctx)
+	sse := subscriber.NewSSESubscriber(environmentRequest.Spec.Config.EtosSse)
+	etrInstance := e.Spec.Instructions.Environment["ENVIRONMENT_ID"]
+
+	logger.Info("Waiting for test runner to start",
+		"environment_request", environmentRequest.Name,
+		"execution_space", e.Name,
+	)
+	for event := range sse.Events(ctx, environmentRequest.Spec.Identifier,
```
andmat900: Can't the iterator return an err? 

```
for event, err := range ...
```
so that error can be logged?

pkg/messaging/subscriber/subscriber.go
```diff
@@ -0,0 +1,142 @@
...
+	logger := logging.FromContextOrDiscard(ctx)
+	url := fmt.Sprintf("%s/%s%s", c.baseUrl, id, filtersToQuery(filter))
+	logger.Info(fmt.Sprintf("Connecting to SSE stream, %s", url))
+	request, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+	if err != nil {
+		return nil, err
+	}
+	client := &http.Client{}
+	response, err := client.Do(request)
+	if err != nil {
```
andmat900: When is err != nil? When server fails to respond (DNS failure, connection refused etc) or also when it responds with 404, 500, 503 etc?

Is additional status code check needed?

```
response, err := client.Do(request)
if err != nil {
    return nil, err
}
if response.StatusCode != http.StatusOK {
    response.Body.Close()
    return nil, fmt.Errorf("unexpected status code connecting to SSE stream ...")
}
return response.Body, nil
```

pkg/provider/executionspace.go
andmat900: Is some context deadline needed here?  The context here seems to come from RunExecutionSpaceProvider  which is just context.TODO(),  no deadline/cancellation.

```
ctx, cancel := context.WithTimeout(ctx, someTimeout)
defer cancel()
```

For cases when the job gets stuck in Pending, or when ETR crashes before sending any event?

andmat900 (changes requested):

pkg/provider/executionspace.go
t-persson: We set a timeout based on the deadline on the environmentrequest https://github.com/eiffel-community/etos/blob/ef224b922b02eb9f948e4118d360e18b46622eac/pkg/provider/provider.go#L312-L314

pkg/messaging/subscriber/subscriber.go
```diff
@@ -0,0 +1,162 @@
...
+	request, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+	if err != nil {
+		return nil, err
+	}
+	client := &http.Client{}
+	var response *http.Response
+	err = retry.Constant(ctx, 5*time.Second, func(ctx context.Context) error {
+		response, err = client.Do(request)
+		if err != nil {
+			return err
```
andmat900, line 106: When client.Do returns a transport error (for example, connection refused during startup), response is nil. The retry loop will treat it as non-retryable and exit. Then the response.Body dereferences a nil -> panic.

andmat900 (changes requested):

pkg/provider/executionspace.go
t-persson: Don't we already check the error for logging?

t-persson: Gah sorry. My review tool is broken

t-persson:
e2e tests fixed by eiffel-community/etos-library#76 . Waiting for it to be released.

pkg/messaging/subscriber/subscriber.go
```diff
@@ -0,0 +1,168 @@
...
+			if closeErr := stream.Close(); closeErr != nil {
+				logger.Error(closeErr, "Error closing stream")
+			}
+		}()
+
+		decoder := sse.NewDecoder(stream)
+		for {
+			event, err := c.decode(decoder)
+			if err != nil {
+				logger.Error(err, "Error decoding event")
```
andmat900: io.EOF will be logged as error

cmd/executionspaceprovider/main.go
```diff
@@ -185,6 +185,16 @@ func (p *genericExecutionSpaceProvider) createExecutionSpaces(
 			span.SetStatus(codes.Error, "failed to start ETOS test runner")
 			return err
 		}
+		if err := executionSpace.WaitForTestRunner(ctx, cfg.EnvironmentRequest); err != nil {
```
andmat900: `WaitForTestRunner` is inside `for range cfg.MinimumAmount`, i. e. the launch becomes serial with this change. This will add start time with multiple subsuites.

andmat900 (changes requested):

pkg/messaging/subscriber/subscriber.go
t-persson: Correct. I don't mind that extra information

pkg/provider/executionspace.go
```diff
@@ -151,6 +154,49 @@ func (e *ExecutionSpace) Create(ctx context.Context) error {
...
+		subscriber.Filter{EventType: events.ShutdownType},
+	) {
+		if err != nil {
+			logger.Error(err, "Error while waiting for test runner to start")
+			return err
+		}
+		switch e := event.(type) {
+		case events.Status:
+			if e.Data.Instance != etrInstance {
+				logger.Info("Received status event for a different ETR instance, ignoring",
```
fredjn: Isn't there a potential that this is going to generate a lot of log spam about things we don't care about? Debug?

t-persson: I will change to debug, but to clarify this would only be printed on each ETR instance for a specific testrun not all ETR instances in ETOS

pkg/messaging/subscriber/subscriber.go
t-persson: Clarified with a stream closed message instead.

cmd/executionspaceprovider/main.go
t-persson: Should be fixed now. I verified that it works with happy-path locally. May verify sad-paths tomorrow.

pkg/provider/executionspace.go
t-persson: Changed

cmd/executionspaceprovider/main.go
t-persson: Sad-paths validated in a local cluster using modified test runners to simulate several error-paths.
I had to fix a few problems with the implementation and I added tests

fredjn (approved):

Issue #661, opened 2026-05-08
Promote v1alpha to v1beta


t-persson:
### Description

We should promote the API up to v1beta when the following issues are solved

- #570 

- [ ] Add v1beta1 version of TestRun
  - [ ] #712 
- [ ] Add v1beta1 version of Environment
  - [ ] #713 
- [ ] Update ETOS library to support v1beta1
  - [ ] eiffel-community/etos-library#68
- [ ] Create a v1beta1 version to ETOS API, which can read the new YAML format
  - [ ] eiffel-community/etos-api#167
  - [ ]  #692 
- [ ] Update Suite Runner to read v1beta1 instead of v1alpha1
- [ ] Update Test Runner to support v1beta1 version of Environment
  - [ ] eiffel-community/etos-test-runner#95
- [ ] Update providers to use v1beta1 of Environment
  - [ ] #714 
- [ ] Update TestRun controller to read v1beta1 instead of v1alpha1
- [ ] Update ETOS client to use ETOS API v1beta1
- [ ] Remove v1alpha1 version of TestRuns

PR #712, opened 2026-06-15
Add v1beta1 version of the TestRun resource

t-persson:
### Applicable Issues

#661 

### Description of the Change

Added v1beta1 version
Added conversion webhook with v1alpha1 as Hub.
The Hub is the version that is stored in ETCD whereas the spoke is a version that can be converted to the Hub version.
This way we will be able to continue to use v1alpha1 within the controller even though v1beta1 versions are created in Kubernetes. Added no other webhooks to v1beta1 since the v1alpha1 webhooks are launched when the v1beta1 is converted with the conversion webhook. When removing v1alpha1 we need to set the Hub to v1beta1 and convert all controllers to support that version instead and we need to add validating and mutating webhooks for v1beta1

### Alternate Designs

I could set v1beta1 as Hub instead of v1alpha1 (means that v1beta1 would be stored in ETCD) but I figured that it would be better to keep the version that we know works and change it in the future when we want to remove/deprecate v1alpha1.
See the steps I added in the issue #661 

### Possible Drawbacks

By keeping both versions we need to maintain the conversion webhook, but that is to be expected and it is good that we get practice on this now.
There are a few parameters in the Spec that I kinda want to get rid off but could not right now. For example the 'identity' is such a parameter, but it is used by both the ETR and the ESR and I could not find a good way to generate an identity in the conversion webhook.

### Sign-off

    indicated in the file; or

    in the file; or

    it.


t-persson:
While we are at it, we should probably remove the LogListener from v1beta1

api/v1beta1/common_conversion.go
```diff
@@ -0,0 +1,143 @@
...
+	dst.Version = src.Version
+	dst.Tracker = src.Repository
+	dst.URI = src.URI
+}
+
+// convertFrom converts the Execution from the v1alpha1 Test to the v1beta1 Execution.
+func (dst *Execution) convertFrom(src *etosv1alpha1.Execution) {
+	var command strings.Builder
+	command.WriteString(src.Command)
+	for key, param := range src.Parameters {
```
andmat900, line 100: The map iteration is not deterministic. Same object may return different command values on consecutive calls.

api/v1beta1/common_conversion.go
```diff
@@ -0,0 +1,143 @@
...
+		if param == "" {
+			command.WriteString(key)
+		} else {
+			fmt.Fprintf(&command, "%s=%s", key, param)
```
andmat900, line 104: There shall be a whitespace separator before the first parameter.

api/v1beta1/testrun_conversion.go
```diff
@@ -0,0 +1,197 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1beta1
+
+import (
+	"fmt"
+	"log"
```
andmat900, line 21: The rest of the codebase uses a different logger.

api/v1beta1/testrun_conversion.go
```diff
@@ -0,0 +1,197 @@
...
+	dst.Spec.Suites = make([]Suite, len(src.Spec.Suites))
+	for i, suite := range src.Spec.Suites {
+
+		testSuite := Suite{}
+		testSuite.convertFrom(&suite)
+		dst.Spec.Suites[i] = testSuite
+	}
+
+	dst.Spec.SuiteRunner = &Image{}
+	if err := dst.Spec.SuiteRunner.convertFrom(src.Spec.SuiteRunner.Image); err != nil {
```
andmat900, line 133: nil checks missing, whereas ConvertTo has nil checks. Should be symmetric.

internal/controller/testrun_controller.go
```diff
@@ -847,14 +848,6 @@ func (r TestRunReconciler) suiteRunnerJob(ctx context.Context, obj client.Object
 			Name:  "SUITE_SOURCE",
 			Value: testrun.Spec.SuiteSource,
 		},
-		{
```
andmat900, line 850: Removal of code in testrun_controller.go seems like a more significant change than promotion to v1beta1.

api/v1beta1/suite.go
```diff
@@ -0,0 +1,161 @@
...
+// TestEnvironment defines configuration for a test execution, including the test runner image, environment variables,
+// and additional provider-defined resources.
+type TestEnvironment struct {
+
+	// TestRunner defines the container image to use as the test runner, e.g.
+	// 'ghcr.io/eiffel-community/etos-base-test-runner:latest'.
+	// +required
+	TestRunner string `json:"testRunner"`
+
+	// EnvironmentVariables defines keyt-value pairs of environment variables to set in the test runner.
```
andmat900, line 140: typo: key-value

api/v1beta1/suite.go
```diff
@@ -0,0 +1,161 @@
...
+	// +kubebuilder:validation:MinLength=1
+	SchemaVersion string `json:"schemaVersion,omitempty"`
+
+	// Suites defines the list of test suites to run.
+	// +required
+	// +kubebuilder:validation:MinItems=1
+	Suites []Suite `json:"suites"`
+}
+
+// Suite defines a single test suite contining prioritized test executions.
```
andmat900, line 42: typo: containing

api/v1beta1/suite.go
```diff
@@ -0,0 +1,161 @@
...
+	// 'ghcr.io/eiffel-community/etos-base-test-runner:latest'.
+	// +required
+	TestRunner string `json:"testRunner"`
+
+	// EnvironmentVariables defines keyt-value pairs of environment variables to set in the test runner.
+	// +optional
+	EnvironmentVariables map[string]string `json:"environmentVariables"`
+
+	// AdditionalResources defines provider-defined resources required by the test beyond the primary IUT (Item Under
+	// Test), e.g. a sidecar containers or additional hardware. Each items fields are open-ended and interpreted by
```
andmat900, line 145: typo: Each item's fields

andmat900, line 145: typo: "a sidecar containers" -> "sidecar containers"

api/v1beta1/testrun_types.go
```diff
@@ -0,0 +1,194 @@
...
+	ID string `json:"id,omitempty"`
+
+	// Artifact defines the unique identifier for the artifact under test as UUID. This field is required.
+	// +required
+	// +kubebuilder:validation:Pattern="^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$"
+	Artifact string `json:"artifact"`
+
+	// Identity defines the identity of the artifact as a packageurl.
+	// +required
+	// +kubebuilder:validation:Pattern="^pkg:[a-z]+/.+$"
```
andmat900: this pattern should allow hyphens and digits instead of just `[a-z]`

andmat900 (changes requested):

api/v1beta1/common_conversion.go
t-persson: I don't think we have ever been deterministic with that field since the src (v1alpha1) is a map and we converted that map to JSON. We can still make it deterministic in v1beta by sorting on the keys, but is this something that is necessary or something we want?

internal/controller/testrun_controller.go
t-persson: Reverted commit, will create separate PR later.

api/v1beta1/common_conversion.go
andmat900: Ignore it if you want.

The problem is that two consecutive `kubectl get testrun -o yaml` calls on an unchanged object can return different command values, which is confusing.

A fix is cheap:
```
for _, key := range slices.Sorted(maps.Keys(src.Parameters)) {
    param := src.Parameters[key]
    command.WriteString(" ")
    if param == "" {
        command.WriteString(key)
    } else {
        fmt.Fprintf(&command, "%s=%s", key, param)
    }
}
```

api/v1beta1/testrun_conversion.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+
+	dst.Spec.SuiteRunner = &Image{}
+	if src.Spec.SuiteRunner != nil {
+		if err := dst.Spec.SuiteRunner.convertFrom(src.Spec.SuiteRunner.Image); err != nil {
+			return err
+		}
+	}
+	dst.Spec.LogListener = &Image{}
+	if err := dst.Spec.LogListener.convertFrom(src.Spec.LogListener.Image); err != nil {
```
andmat900: nil guard missing. src.Spec.LogListener is *LogListener with omitempty, so it can be nil.

andmat900 (changes requested):

andmat900 (approved):

fredjn (approved):

PR #713, opened 2026-06-16
Add v1beta1 version of the Environment resource

t-persson:
### Applicable Issues

#661 
shares common types with #712 

### Description of the Change

The v1beta1 version will use the suite structure of the v1beta1 testrun being created in another pull request.
Added conversion webhook with v1alpha1 as Hub.
The Hub is the version that is stored in ETCD whereas the spoke is a version that can be converted to and from the Hub version. This way we will be able to continue to use v1alpha1 within the controller even though v1beta1 versions are created in Kubernetes. When removing v1alpha1 we need to set the Hub to v1beta1 and convert all controllers to support that version instead.

### Alternate Designs

I could set v1beta1 as Hub instead of v1alpha1 (means that v1beta1 would be stored in ETCD) but I figured that it would be better to keep the version that we know works and change it in the future when we want to remove/deprecate v1alpha1.
See the steps I added in the issue https://github.com/eiffel-community/etos/issues/661

I kinda wanted to remove the iut, executor and logArea in favor of the respective resources, but that change would be really a lot bigger than I think we want and it would also add another three requests to the Kubernetes API for each subsuite we get (since the ETOS API would need to get all resources for the Environment).

### Possible Drawbacks

By keeping both versions we need to maintain the conversion webhook, but that is to be expected and it is good that we get practice on this now.

### Sign-off

    indicated in the file; or

    in the file; or

    it.


api/v1beta1/environment_types.go
```diff
@@ -0,0 +1,153 @@
...
+	// +required
+	Context string `json:"context"`
+
+	// TestRunner defines the image to use for the test runner.
+	// +kubebuilder:validation:MinLength=1
+	// +required
+	TestRunner string `json:"testRunner"`
+
+	// Providers defines which providers were used to create this environment
+	// +required
```
andmat900, line 66: +required and omitempty is contradictory

api/v1alpha1/environment_conversion.go
```diff
@@ -0,0 +1,22 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1alpha1
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
```
andmat900, line 19: Leftover scaffolding?

api/v1beta1/common_conversion.go
```diff
@@ -0,0 +1,118 @@
...
+	dst.Version = src.Version
+	dst.Tracker = src.Repository
+	dst.URI = src.URI
+}
+
+// convertFrom converts the Execution from the v1alpha1 Test to the v1beta1 Execution.
+func (dst *Execution) convertFrom(src *etosv1alpha1.Execution) {
+	var command strings.Builder
+	command.WriteString(src.Command)
+	for key, param := range src.Parameters {
```
andmat900, line 75: The map iteration is not deterministic. Same object may return different command values on consecutive calls.

api/v1beta1/common_conversion.go
```diff
@@ -0,0 +1,118 @@
...
+		if param == "" {
+			command.WriteString(key)
+		} else {
+			fmt.Fprintf(&command, "%s=%s", key, param)
```
andmat900, line 79: There shall be a whitespace separator before the first parameter.

api/v1beta1/environment_conversion.go
```diff
@@ -0,0 +1,111 @@
...
+// environmentLog is for logging Environment conversions.
+var environmentLog = logf.Log.WithName("environment-conversion")
+
+// ConvertTo converts this Environment (v1beta1) to the Hub version (v1alpha1).
+func (src *Environment) ConvertTo(dstRaw conversion.Hub) error {
+	dst := dstRaw.(*etosv1alpha1.Environment)
+	environmentLog.Info(fmt.Sprintf("ConvertTo: Converting Environment from Spoke version v1beta1 to Hub version v1alpha1;"+
+		"source: %s/%s, target: %s/%s", src.Namespace, src.Name, dst.Namespace, dst.Name))
+
+	dst.Spec.Name = src.Spec.Name
```
andmat900: should Dataset be included here?

api/v1beta1/common_conversion.go
```diff
@@ -13,7 +13,6 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-
 package v1beta1
```
andmat900: Accidental edit? With no blank line, Go treats the Apache license block as the package doc comment for v1beta1.

api/v1beta1/environment_conversion.go
```diff
@@ -0,0 +1,111 @@
...
+	dst.Spec.ID = src.Spec.SubSuiteID
+	dst.Spec.MainSuiteID = src.Spec.MainSuiteID
+	dst.Spec.TestrunID = src.Spec.SuiteID
+	dst.Spec.Artifact = src.Spec.Artifact
+	dst.Spec.Context = src.Spec.Context
+	dst.Spec.TestRunner = src.Spec.TestRunner
+	dst.Spec.Deadline = src.Spec.Deadline
+	dst.Spec.Providers = Providers{
+		IUT:            src.Spec.Providers.IUT,
```
andmat900: nil guard needed? It can be nil in v1alpha.

andmat900 (changes requested):

api/v1beta1/environment_conversion.go
t-persson: The v1alpha1 `EnvironmentSpec` has no Dataset field, so the v1beta1 Dataset cannot be copied during conversion without changing the hub API and its consumers.

andmat900 (approved):

